Issue #15, opened 2020-04-14
Error reading fort.44 file from recent SOLPS


jacklovell:
The fort.44 loading function errors out when reading files produced by a recent version of SOLPS-ITER. The format identifier for these files is 20170328.

The Eirene reading logic has been tested on a file with format ID 20130210 (AUG test case), and I've also successfully read an Eirene file with format ID 20070912. But 20000727 does not work.

We should explicitly check the Eirene format ID, and raise an error for unsupported versions until new routines to read these format IDs are added.

Mateasek:
I agree, last week I tried to read file with format ID 20170328 (first time I had to work with this module) which has comments before data blocks. This is not expected by the parsing script.

I think it would be in general a better approach to pull the file parsing code outside of Eirene class and make it accept simple np arrays as it is in the case of the EFITEquilibrium class. Then the parsing could be done in methods which can either return data in a dictionary or instance of the Eirene class directly. This way there will be simple way how to add parsing for more versions.

jacklovell:
I have an in-progress branch which is able to read the 20170328 format of this. Like @Mateasek suggests, I've split the 2013 and 2017 formats into separate functions, with a top level function whose only job is to read the format ID and delegate to the appropriate function. I'll clean this up and submit it as a PR for review.

Mateasek:
Awesome, thanks @jacklovell. Tell me if you need any help.

Mateasek:
I tested the funciton reading 2017 fort.44 file version in the #16. I anyway think that the parsing should be moved outside of the eirene class, which would become only a data containter with defined inputs, variables and properties. There are two main reasons.

 First is it will give the possibility to move the file reading code into separate files which will make the maintenance easier in my opinion. Managing of functions and assigning eirene versions in case a function covers mutiple would be easy and a new function or class can be added any time a new incompatible eirene version comes up.

Second is that this solps package in my opinion should handle only data formats exported by solps and other formats specific to institutes or machies should be handled externally. They could provide neccessary data in a specified format to the solps classes from this repository like it is in the case of EFITEquilibrium class. This would in my opinion also make the maintenance easier. An example is that for my case I would create a "solps data getter" class for COMPASS in the compass repository using classes from this repository.

jacklovell:
[...]

PR #16, opened 2020-04-21
Add support for reading 2017 fort.44 files.

jacklovell:
The fort.44 file format changes between SOLPS releases. This commit
adds support for the 2017 format, and also explicitly checks the format
ID of the file to ensure it is in a format Cherab understands.

### DIFF
--- a/cherab/solps/eirene/eirene.py
+++ b/cherab/solps/eirene/eirene.py
@@ -1,4 +1,3 @@
-
 # Copyright 2016-2018 Euratom
 # Copyright 2016-2018 United Kingdom Atomic Energy Authority
 # Copyright 2016-2018 Centro de Investigaciones Energéticas, Medioambientales y Tecnológicas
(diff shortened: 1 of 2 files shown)


jacklovell:
@Mateasek could you please test this with the files you found were failing (as you mentioned in #15) to see if it fixes that problem?

mattngc (approved):
Nice work, looks fine to me. Obviously important to here @Mateasek's feedback on the tests.

Do you have any ideas how many legitimate SOLPS / Eirene formats there are? If there are a lot then it may be easier to redesign this and simply have a class per valid format. Overriding the base functions where necessary, etc...

jacklovell:
I've come across at least 4 formats so far, but I don't know how different they are: the format ID appears to be tied to particular versions of Eirene which are then tied to particular versions of SOLPS.

I think it's valuable to have a single class which abstracts away the fort.44 format, and an automatic dispatch to the file reading routines once the format of the file is detected. This will make user code significantly simpler, and more reusable across different SOLPS runs. Cherab is only using a subset of the data stored in the fort.44 file, so the `Eirene` class isn't that tightly coupled to the file format to justify different classes for different file formats, I think.

Mateasek (approved):
I tested it on our SOLPS output with  20170328  3.0.6-81-ga3fccb2              
 fort.44 version and it works flawlessly, thanks for adding the parsing function for the new version @jacklovell

PR #17, opened 2020-04-21
Create balance.py

jrh-ccfe:
This code creates a plasma object from balance.nc files written by SOLPS-ITER


cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the Licence is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied.
+#
+# See the Licence for the specific language governing permissions and limitations
+# under the Licence.
+
+import os
+import copy
+from numpy import abs,sum,swapaxes,ones,arange
```
jacklovell: It's generally preferable to do `import numpy as np` when you're using `abs` and `sum`, as these overwrite Python's built-in functions. They appear much further down the file too, by which time a reader may have forgotten or missed that they're imported from numpy and wonder why you're passing the `axis` keyword when the Python functions don't support that. Could you please change these calls to `np.sum`, `np.abs` etc for clarity?

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+
+from cherab.core.math.mappers import AxisymmetricMapper
+from cherab.core.atomic.elements import hydrogen, deuterium, helium, beryllium, carbon, nitrogen, oxygen, neon, \
+    argon, krypton, xenon
+
+from cherab.solps.eirene import Eirene
+from cherab.solps.mesh_geometry import SOLPSMesh
+from cherab.solps.solps_plasma import SOLPSSimulation
+
+Q = 1.602E-19
```
jacklovell: Could you use `scipy.constants.elementary_charge` instead please? This is also used internally by Cherab core. The fewer hard coded, slightly different values of e that we have scattered through the code base the better.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+    (10, 20.180): neon,
+    (18, 39.948): argon,
+    (18, 40.0): argon,
+    (36, 83.798): krypton,
+    (54, 131.293): xenon
+}
+
+
+# Code developed by J. Harrison 9/4/2019
+def load_solps_from_balance(balance_filename, debug=False):
```
jacklovell: You don't currently use the `debug` argument. Either remove it, or add some `if debug:` statements if you want to print out debugging information.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+    """
+    Load a SOLPS simulation from SOLPS balance.nc output files.
+
+    """
+
+    if not os.path.isdir(balance_filename):
+        RuntimeError("file name must be valid")
```
jacklovell: This block has no effect: an exception will only be raised if you use the `raise` statement. Currently you're just creating an instance of an exception and then not using it.

In any case, this block is redundant. If the file name does not point to a valid netCDF file then line 61 will raise a `FileNotFoundError` anyway. So this block can be removed.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+    fhandle = netcdf.netcdf_file(balance_filename,'r')
+	
+    # Load SOLPS mesh geometry
+    cr_x = copy.deepcopy(fhandle.variables['crx'].data)
+    cr_z = copy.deepcopy(fhandle.variables['cry'].data)
+    vol = copy.deepcopy(fhandle.variables['vol'].data)
+	
+    # Re-arrange the array dimensions in the way CHERAB expects...
+    cr_x = swapaxes(cr_x,0,2)
+    cr_x = swapaxes(cr_x,0,1)
```
jacklovell: You can accomplish this in a single line. Since you're really moving the first dimension to the end, you can do:
```python
cr_x = np.moveaxis(cr_x, 0, -1)
```
I personally find this version better conveys your actual intent.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+    cr_z = swapaxes(cr_z,0,2)
+    cr_z = swapaxes(cr_z,0,1)
```
jacklovell: Same comment re: moveaxis.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+    if not os.path.isdir(balance_filename):
+        RuntimeError("file name must be valid")
+    
+    # Open the file
+    fhandle = netcdf.netcdf_file(balance_filename,'r')
+	
+    # Load SOLPS mesh geometry
+    cr_x = copy.deepcopy(fhandle.variables['crx'].data)
+    cr_z = copy.deepcopy(fhandle.variables['cry'].data)
+    vol = copy.deepcopy(fhandle.variables['vol'].data)
```
jacklovell: `deepcopy` is unnecessary here: you can use the numpy array's `copy` method instead to create a new array which owns its data. It's neater and faster, and saves an import:
```python
cr_x = fhandle.variables['crx'].data.copy()
```

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+
+    sim = SOLPSSimulation(mesh)
+    ni = mesh.nx
+    nj = mesh.ny
+
+    # TODO: add code to load SOLPS velocities and magnetic field from files
+
+    # Load electron species
+    sim._electron_temperature = copy.deepcopy(fhandle.variables['te'].data)/Q
+    sim._electron_density = copy.deepcopy(fhandle.variables['ne'].data)
```
jacklovell: Same comment as above re: deepcopy.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+        if fhandle.variables['species'].data[i,1] == b'N' and fhandle.variables['species'].data[i,2] == b'e':
+            zn = 10
+        if fhandle.variables['species'].data[i,1] == b'A' and fhandle.variables['species'].data[i,2] == b'r':
+            zn = 18
+	
+        am = float(fhandle.variables['am'].data[i])  # Atomic mass
+        charge = int(fhandle.variables['za'].data[i])  # Ionisation/charge
+        species = _popular_species[(zn, am)]
+		
+        # If we only need to populate species_list, there is probably a faster way to do this...
```
jacklovell, line 100: Since the number of species is typically small, you probably don't need to worry about performance. Your intent here is clear: that's more important.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+            zn = 18
+	
+        am = float(fhandle.variables['am'].data[i])  # Atomic mass
+        charge = int(fhandle.variables['za'].data[i])  # Ionisation/charge
+        species = _popular_species[(zn, am)]
+		
+        # If we only need to populate species_list, there is probably a faster way to do this...		
+        sim.species_list.append(species.symbol + str(charge))
+
+    tmp = copy.deepcopy(fhandle.variables['na'].data)
```
jacklovell: Same comment as above re: deepcopy.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+    tmp = swapaxes(tmp,0,2)
+    tmp = swapaxes(tmp,0,1)
+    sim._species_density = tmp
```
jacklovell: Same comment as above re: moveaxis.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+            if sim.species_list[i] == 'D0':
+                D0_indx = i
+
+    # Replace the deuterium neutrals density (from the fluid neutrals model by default) with
+    # the values calculated by EIRENE - do the same for other neutrals?
+    if 'dab2' in fhandle.variables.keys():
+        sim.species_density[:,:,D0_indx] = copy.deepcopy(fhandle.variables['dab2'].data[0,:,0:-2])
+        eirene_run = True
+    else:
+        eirene_run = False
```
jacklovell: There are a few issues here:
1. `D0_indx` is only defined if D0 appears in `sim.species_list`. If the balance file somehow has a `dab2` variable but no D0 species then `D0_indx` in line 134 will be undefined and you'll get an `UnboundLocalError` exception raised. For robustness, you could set `D0_indx` to a sentinel value such as `None` beforehand, and only update the neutral density if `D0_indx` is not `None`.
2. Lines 125-129 can be rewritten as: `D0_indx = sim.species_list.index('D0')`. Python has batteries included!
3. Same comment about deepcopy.

You can re-write this section more compactly as the following:
```python
try:
    D0_indx = sim.species_list.index('D0')
except ValueError:  # No 'D0'
    eirene_run = False
    # Could instead raise an error here, if a lack of D0 is a critical error.
else:
    if 'dab2' in fhandle.variables.keys():
        sim.species_density[:, :, D0_indx] = fhandle.variables['dab2'].data[0, :, 0:-2].copy()
        eirene_run = True
    else:
        eirene_run = False
```

I also ran into an error in this part when testing against a balance.nc file I have: 
```python
In [59]: sim = load_solps_from_balance('/home/jlovell/Bolometer/SXD phantoms/Full data/seed_8.nc')       
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-59-de7e0aeae2c8> in <module>
----> 1 sim = load_solps_from_balance('/home/jlovell/Bolometer/SXD phantoms/Full data/seed_8.nc')

~/cherab/cherab_solps/cherab/solps/formats/balance.py in load_solps_from_balance(balance_filename, debug)
    132     # the values calculated by EIRENE - do the same for other neutrals?
    133     if 'dab2' in fhandle.variables.keys():
--> 134         sim.species_density[:,:,D0_indx] = copy.deepcopy(fhandle.variables['dab2'].data[0,:,0:-2])
    135         eirene_run = True
    136     else:

ValueError: could not broadcast input array from shape (34,151) into shape (34,148)
```
Looks like there is a dimension issue happening somewhere.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+    
+    # Calculate the total radiated power
+    
+    
+    if eirene_run is True:
```
jacklovell: `if eirene_run:` is sufficient: no need to compare identity to `True`.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+        # Electron energy loss due to interactions with neutrals
+        if 'eirene_mc_eael_she_bal' in fhandle.variables.keys():
+            eirene_ecoolrate = sum(fhandle.variables['eirene_mc_eael_she_bal'].data,axis=0)/vol
+            
+        # Ionisation rate from EIRENE, needed to calculate the energy loss to overcome the ionisation potential of atoms
+        if 'eirene_mc_papl_sna_bal' in fhandle.variables.keys():
+            eirene_potential_loss = 13.6*sum(fhandle.variables['eirene_mc_papl_sna_bal'].data,axis=(0))[1,:,:]*Q/vol
+        
+        # This will be negative (energy sink); take absolute valu
+        sim._total_rad = abs(b2_ploss+(eirene_ecoolrate-eirene_potential_loss))
```
jacklovell: Is it the absolute value we want here? Or do we strictly want to return the negative of the energy loss for the radiated power? Semantically I think the latter, but is there ever the possibility of the energy loss term returned by Eirene being positive (which would correspond to negative radiated power)?

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+        
+    else:
+        #print(b2_ploss.shape)
+        #print(cr_x.shape)
+        #print(sum(b2_ploss,axis=0).shape)
```
jacklovell: Please remove commented out code. Or, wrap it in an `if debug:` block and make use of your debug argument.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+      Second dictionary has a ndarray for each piece of data found in the file.
+    """
+    _, _, geom_data_dict = load_b2f_file(mesh_file_path, debug=debug)
+
+    cr_x = geom_data_dict['crx']
+    cr_z = geom_data_dict['cry']
+    vol = geom_data_dict['vol']
+
+    # build mesh object
+    return SOLPSMesh(cr_x, cr_z, vol)
```
jacklovell: This is identical to a function with the same name in `raw_simulation_files.py`. Overzealous copy and paste?

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,192 @@
...
+        #print(b2_ploss.shape)
+        #print(cr_x.shape)
+        #print(sum(b2_ploss,axis=0).shape)
+        
+        # Total radiated power from B2, not including neutrals
+        b2_ploss = sum(fhandle.variables['b2stel_she_bal'].data,axis=0)/vol
+        
+        potential_loss = sum(fhandle.variables['b2stel_sna_ion_bal'].data,axis=0)/vol
+        
+        #rad_dens = (-sum(b2stel_she_bal,3)+13.6*Q*b2stel_sna_ion_bal[:,:,0])/vol
```
jacklovell: Again, remove commented out code.

jacklovell (changes requested):
Thanks for this James! It's a very useful addition. I have a few comments which I've added inline for you to address. I also found an error when trying to read a balance.nc file I have in my home directory: see the comment for lines 125-137.

In addition, there are a couple of general formatting issues. Please use spaces after commas everywhere, to keep consistent with [PEP8](https://www.python.org/dev/peps/pep-0008/) and the rest of the Cherab code base. There's a bit of trailing whitespace too: please remove.

cherab/solps/formats/balance.py
jrh-ccfe: Done

cherab/solps/formats/balance.py
jrh-ccfe: Done

cherab/solps/formats/balance.py
jrh-ccfe: Done - this block has been removed

cherab/solps/formats/balance.py
jrh-ccfe: Done

cherab/solps/formats/balance.py
jrh-ccfe: Done

cherab/solps/formats/balance.py
jrh-ccfe: Done, the term in brackets is now multiplied by -1

cherab/solps/formats/balance.py
jrh-ccfe: Done

cherab/solps/formats/balance.py
jrh-ccfe: Done

cherab/solps/formats/balance.py
jrh-ccfe: Exactly right, this function has been removed

cherab/solps/formats/balance.py
jrh-ccfe: Done

cherab/solps/formats/balance.py
jrh-ccfe: Done

cherab/solps/formats/balance.py
jrh-ccfe: Done

cherab/solps/formats/balance.py
jrh-ccfe: Done

cherab/solps/formats/balance.py
jrh-ccfe: Done

cherab/solps/formats/balance.py
jrh-ccfe: Done

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,165 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the Licence is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied.
+#
+# See the Licence for the specific language governing permissions and limitations
+# under the Licence.
+
+import os
+import copy
```
jacklovell: These imports are no longer needed: please remove them.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,165 @@
...
+import numpy as np
+from scipy.io import netcdf
+from scipy.constants import elementary_charge
+from raysect.core.math import Discrete2DMesh
+
+from cherab.core.math.mappers import AxisymmetricMapper
+from cherab.core.atomic.elements import hydrogen, deuterium, helium, beryllium, carbon, nitrogen, oxygen, neon, \
+    argon, krypton, xenon
+
+from cherab.solps.eirene import Eirene
```
jacklovell: This import is unused: please remove.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,165 @@
...
+    # Re-arrange the array dimensions in the way CHERAB expects...
+    cr_x = np.moveaxis(cr_x, 0, -1)
+    cr_z = np.moveaxis(cr_z, 0, -1)
+	
+    # Create the SOLPS mesh
+    mesh = SOLPSMesh(cr_x,cr_z,vol)	
+
+    sim = SOLPSSimulation(mesh)
+    ni = mesh.nx
+    nj = mesh.ny
```
jacklovell: These are unused, and can be removed.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,165 @@
...
+# Code developed by J. Harrison 9/4/2019
+def load_solps_from_balance(balance_filename):
+    """
+    Load a SOLPS simulation from SOLPS balance.nc output files.
+
+    """
+
+    # Open the file
+    fhandle = netcdf.netcdf_file(balance_filename,'r')
+
```
jacklovell: You have trailing whitespace here, along with quite a few other places in the file. Please remove it. A quick way to do this is:
```bash
sed -i 's/[[:blank:]]\+$//g' balance.py
```

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,165 @@
...
+# Code developed by J. Harrison 9/4/2019
+def load_solps_from_balance(balance_filename):
+    """
+    Load a SOLPS simulation from SOLPS balance.nc output files.
+
+    """
+
+    # Open the file
+    fhandle = netcdf.netcdf_file(balance_filename,'r')
```
jacklovell: Per [PEP-8](https://www.python.org/dev/peps/pep-0008/), please add a space after the comma. This applies to quite a few places throughout the file. Running `pylint balance.py` will point out the others.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,165 @@
...
+    mesh = SOLPSMesh(cr_x,cr_z,vol)	
+
+    sim = SOLPSSimulation(mesh)
+    ni = mesh.nx
+    nj = mesh.ny
+
+    # TODO: add code to load SOLPS velocities and magnetic field from files
+
+    # Load electron species
+    sim._electron_temperature = fhandle.variables['te'].data.copy()/Q
```
jacklovell: PEP8 again: please put spaces around the `/` operator.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,165 @@
...
+    sim._species_density = tmp
+
+    # Make Mesh Interpolator function for inside/outside mesh test.
+    inside_outside_data = np.ones(mesh.num_tris)
+    inside_outside = AxisymmetricMapper(Discrete2DMesh(mesh.vertex_coords, mesh.triangles, inside_outside_data, limit=False))
+    sim._inside_mesh = inside_outside
+
+    
+    # Load the neutrals data
+    D0_indx = sim.species_list.index('D0')
```
jacklovell: This will raise a `ValueError` if there is no 'D0' in the species list. Probably safest to wrap this in a try/except block if you want to be able to handle balance.nc files without neutral deuterium:
```python
try:
    D0_indx = sim.species_list.index('D0')
except ValueError:
    D0_indx = None
```
Of course, if missing D0 data is a critical error then there's no point catching this, but then you don't need to do the `if D0_indx is not None` check in line 127.

cherab/solps/formats/balance.py
```diff
@@ -0,0 +1,165 @@
...
+        
+    else:
+ 
+        # Total radiated power from B2, not including neutrals
+        b2_ploss = np.sum(fhandle.variables['b2stel_she_bal'].data,axis=0)/vol
+        
+        potential_loss = np.sum(fhandle.variables['b2stel_sna_ion_bal'].data,axis=0)/vol
+        
+        # Save total radiated power to the simulation object
+        sim._total_rad = 13.6*Q*potential_loss-b2_ploss
```
jacklovell: Please add spaces around the multiplication and subtraction operators here, as well as the division operators in the lines above. Pylint will tell you if you've missed any.

demos/solps_from_balance.py
```diff
@@ -0,0 +1,163 @@
...
+from cherab.core.atomic.elements import carbon, deuterium
+from cherab.solps import load_solps_from_balance
+
+plt.ion()
+
+xl, xu = (0.0, 2.0)
+yl, yu = (-2.0, 2.0)
+
+print('Enter name of balance.nc file:')
+filename = input()
```
jacklovell, line 32: I like that users can specify any balance.nc file here in the demo, but you do need to be careful about the assumptions you make. For example, not all runs will have D0, D1 and C0-C6: I have some C-Mod runs with no carbon. I wouldn't worry about modifying the code to catch all these different possibilities, but I would suggest adding a comment here to say that for this demo to run properly the input file must match the assumptions you're making about its contents.

demos/solps_from_balance.py
```diff
@@ -0,0 +1,163 @@
...
+plt.xlim(xl, xu)
+plt.ylim(yl, yu)
+plt.title("D0 velocity")
+
+plt.figure()
+plt.imshow(d1_velocity, extent=[xl, xu, yl, yu], origin='lower')
+plt.colorbar()
+plt.xlim(xl, xu)
+plt.ylim(yl, yu)
+plt.title("D1 velocity")
```
jacklovell: The D0 and D1 bulk velocities are not defined, so these plots will fail. Delete this section.

jacklovell (changes requested):
Thanks for the updated code. I have only a few more stylistic comments, for consistency with the rest of the code base, and a few smaller suggestions. Once these have been addressed I think this will be ready for merging.

jacklovell (approved):
Thanks for the updates James. Happy to merge this now.

PR #29, opened 2020-07-29
Release 1.1.0

### LINKED ISSUE #21: New release
The current master branch of cherab-solps is incompatible with the released versions of raysect and cherab-core. This means new users need to check out the development branch before installing, which has led to confusion in a number of cases. There have been recent proposals to re-architect parts of this repository (e.g. input file reading), so I think it makes sense to make a new release of the current code before large scale changes are made.

For the release, I think the following is required (updated with additional suggestions):

- [x] Update Changelog.
- [x] Bump version number (1.1.0?)
- [x] Flesh out README with some installation instructions, including the need to use Cython 3.0a instead of 0.29.x for a reliable installation. See https://github.com/cherab/Soledge2D/blob/development/README.md for what I had in mind.
- [ ] Self-contained demo with example SOLPS output files.

@mattngc @CnlPepper do you agree with this?

jacklovell:
This release adds support for the latest released versions of Raysect and Cherab core. Without it, we are in the unfortunate situation that cherab-solps master is incompatible with raysect and cherab-core master, and cherab-solps development is also incompatible with raysect and cherab-core development. Once the release is published we can update the development branch to bring it in line with raysect and cherab-core and still have a collection of working packages.

Closes #21

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
 Project Changelog
 =================
 
+Release 1.1.0 (30 July 2020)
+----------------------------
+
+* Fix mesh loading when SOLPSPlasma is unpickled.
+* Support Raysect 0.6 and Cherab-core 1.2.
+* Add support for multiple versions of Eirene fort.44 files.
+* Add support for loading a simulation from a balance.nc file.
+* Fix MDSplus neutral radiation and density loading.
+* Add boron to the species list when creating a plasma from a SOLPSSimulation.
+
 Release 1.0.0 (13 Sept 2017)
 ----------------------------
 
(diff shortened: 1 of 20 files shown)


jacklovell:
You're right, neither is a particularly appealing solution. At least this way the most common issue we've encountered so far with building the project won't trip everybody up. I've been testing it for the past few months (also with core) and haven't run into any problems yet, but it's never guaranteed. The sooner Cython releases 3.0 stable the sooner we can update the instructions, but I think this is a pragmatic solution for now.

jacklovell:
Are there any other objections to the merge?

Issue #34, opened 2020-09-03
Use more SOLPS data when available


vsnever:
Currently, the following useful data are extracted from the SOLPS files, but not used when creating a plasma object:

- neutral density (from EIRENE, see #31),

- effective neutral temperature (from EIRENE), 

- ion temperature (from B2),

- parallel velocities (from B2),

- magnetic field vector (from b2fgmtry),

- total (molecular and atomic) H-alpha emission (from EIRENE).

The total H-alpha emission is a separate issue, because it may require changing the plasma object. Implementing the usage of other quantities from this list when creating the plasma object should be relatively easy. Currently, for neutral atoms the Maxwellian distribution is created with the electron temperature, which at the edge can be an order of magnitude higher than the effective temperature of atoms. The electron temperature is used also for ions even if the ion temperature is available.

Issue #28, opened 2020-06-12
Inconsistent treatment of ENEUTRAD among input formats


jacklovell:
The routines for reading in the total radiated power from neutrals are inconsistent:
- For raw files, the ENEUTRAD quantity from the fort.44 file is used directly. This gives power in W, is only for neutrals (no impurity ion radiation is counted) and negative values correspond to positive emission.
- For MDSplus, abs(bremsstrahlung + line radiation +  eneutrad) / volume. This gives emissivity in W/m3.
- For balance.nc, depends on whether or not Eirene data is available, but essentially -(b2 power loss + potential loss + [ecool rate if available]) / volume. Again, this gives emissivity in W/m3.

All 3 of these are exposed by `SOLPSSimulation.total_radiation`, and the interpolator `SOLPSSimulation.total_radiation_volume`. So we have the unfortunate situation where the quantity stored in this variable depends on the data source. It would be OK to store the same quantity calculated in different ways (e.g. W/m3 emissivity with different contributions for the 3 reading methods), but I think storing either emissivity or power is a recipe for confusion.

I propose we modify the raw file reading, so that the total power is given by -ENEUTRAD / volume. This would produce a quantity in W/m3, where positive values correspond to emission. This is not only consistent with the MDSplus and balance.nc variants, but also more consistent with the rest of the Cherab framework (the `TotalRadiatedPower` model for example outputs quantities in W/m3/sr, rather than W).

We could also modify the output to read -(ENEUTRAD + EIONRAD + EMOLRAD) / volume, which should be a more accurate measure of total radiation. But the latter two quantities aren't available in fort.44 files in the 2013 format or older.

@mattngc @Mateasek @jrh-ccfe @shenders what are your thoughts on this?

mattngc:
This sounds like a sensible approach to me.

Impressed you managed to track down all this information. Remember I could never get consistency between the different formats, it seemed like every format was different and documentation was rare. I remember doing a benchmark for Felix with one of his cases and we managed to get reasonable agreement... but long since forgotten now.

Mateasek:
[...]

PR #35, opened 2020-09-04
Fixes and improvements in SOLPSSimulation, SOLPSMesh and related functions

### LINKED ISSUE #8: SOLPS rewrite and tidy-up
There is a lot of bad code in this package that really needs addressing. At present the SOPLS sim object has no way of being initialised without modifying internal, hidden attributes. I've spotted a few cases of API inconsistency and incorrect docstrings.

### LINKED ISSUE #20: Add class functions returning data interpolators to SOLPSSimulation
I think it would be a nice addition to the SOLPSSimulation. The interpolators are created within the create_plasma function. Moving the code to separate funcions would add a nice possibility to use the solps module for plotting, data generation and etc. together with cherab samplers.

For such purposes functions returning 2D interpolators would be enough, but adding functions returning 3D interpolators shouldn't harm anyone. I would be happy to do this.

### LINKED ISSUE #31: Plasma from raw files should use Eirene neutral density if available
The `dab2` quantity computed by Eirene should be used for the neutral density. B2 uses a fluid approximation for the density which is not self-consistent with the rest of the simulation, and is typically orders of magnitude too low. Both MDSplus and balance readers use dab2 for neutrals already.

### LINKED ISSUE #33: Look up in cherab.core.atomic.elements instead of using _popular_species and _species_symbol_map dicts
I suggest that instead of using _popular_species dict when reading the data from raw or balance files, we should search for the matches in `cherab.core.atomic.elements`. First, get all isotopes with a given nuclear charge, second, from them find the one with the closest atomic mass to the given one.

Similarly, `_species_symbol_map` dict must be replaced with the search for the matching element symbol in `cherab.core.atomic.elements`. Extracting the nuclear charge from the balance file also must be improved the same way.

### LINKED ISSUE #36: Fix velocity calculation
Currently, the parallel velocity (`UA` in SOLPS notation) is misinterpreted as the projection of velocity onto the poloidal direction. This is not so, parallel velocity is the projection of velocity onto the direction of the magnetic field: V<sub>||</sub> = (**VB**)/|**B**|.

To correctly calculate the velocities we must use particle fluxes at poloidal and radial directions defined on cell faces. This will give us 4 velocity vectors for each cell at the cell faces. By averaging them we'll get the projection of velocity on the poloidal plane for each cell, **V**<sub>plane</sub>. Then using **V**<sub>plane</sub>, V<sub>||</sub> and **B**, we can obtain V<sub>tor</sub> (many thank to Andrey Pshenov for clarifying all this).

Note that if the flux from cell A to cell B is positive, than the velocity at the face between A and B is obtained by dividing this flux by the density at cell A, but if it's negative, than the flux must be divided by the density at cell B.

I'll add this to the code.

### LINKED ISSUE #37: Invert index order of SOLPS data arrays or make all arrays as column-major
The current index order of SOLPS data arrays: (poloidal, radial, species) make sense only for column-major arrays (Fortran-style or F-style), because the species slice is needed more often than the radial slice and the radial slice is needed more often than the poloidal slice.

In the raw simulation files, the arrays are stored in column-major order, and `load_b2f_file()` and `read_block44()` return F-style arrays. But on the MDS server they are stored in row-major order (C-style) and with inverse indexing. When reading from MDS, the index order is inverted but the order is not changed from C-style to F-style.

Thus, to improve the performance we should either change the order from C-style to F-style when reading from MDS and also make sure that all arrays created for operation with the SOLPS data arrays are F-style, or to make all arrays C-style and invert the indexing: (species, radial, poloidal). I like the second option better because C-style arrays are default in numpy.

vsnever:
This PR addresses the issues #8, #20, #31, #33, #34, #36 and #37.

Issue #8 is addressed in the following way:

- Safe to use `SOLPSFunction2D` and `SOLPSVectorFunction2D` classes are implemented. They replace  `SOLPSFunction3D` and `SOLPSVectorFunction3D`. For 3D, use `AxisymmetricMapper(SOLPSFunction2D)` and `VectorAxisymmetricMapper(SOLPSVectorFunction2D)`.

- `SOLPSMesh` and `SOLPSSimulation` can now be initialised without modification of hidden attributes;

- `SOLPSMesh` has new attributes for basis vectors, cell connection areas, indices of neighbouring cells and new functions `to_poloidal()` and `to_cartesian()` for converting vectors defined on a grid from/to (poloidal, radial)/(R, Z);

- `species_list` in `SOLPSSimulation` is a list of (name, charge) pairs now instead of a list of strings;

- Setting `b_field` and `velocities` vectors in curvilinear poloidal coordinates (**e**<sub>pol</sub>, **e**<sub>rad</sub>, **e**<sub>tor</sub>) automaticaly sets `b_field_cylindrical` and `velocities_cylindrical` (**e**<sub>R</sub>, **e**<sub>φ</sub>, **e**<sub>Z</sub>) and vice versa (note: **e**<sub>tor</sub> = -**e**<sub>φ</sub>);

- Incorrect calculation of basis vectors using cell centres is fixed (basis vectors are defined by left and bottom faces of the cell);

- The code for `mesh.triangles` and `mesh.triangle_to_grid_map` creation is vectorised (as all other code wherever possible);

- `SOLPSSimulation` now has setters for its properties;

- `load_solps_from_mdsplus()`, `load_solps_from_raw_output()` and `load_solps_from_balance()` do not brake `SOLPSSimulation` interface anymore;
   
- `load_solps_from_mdsplus()` and `load_solps_from_raw_output()` are made more robust. Now both support B2 standalone runs. Error handling is added to `load_solps_from_mdsplus()` for some optional attributes that may not be present on the server;

- Other small improvements.

Issue #20 is fixed. `*_f2d()` and `*_f3d()` interpolators are added for plasma parameters. For vectors `_carteisan` suffix is used instead of `_f3d`.

Issue #31 is fixed. Now if the neutral densities from Eirene are available, they replace the neutral densities from B2.

Issue #33 is fixed using `lookup_isotope()` and `lookup_element()` methods. The plasma composition is defined from nuclear charge, atomic mass number and electrical charge. This is more robust than parsing a string with species names.

Issue #34 is addressed by parsing the following additional quantities in `load_solps_from_mdsplus()` and `load_solps_from_raw_output()`:

- Ion temperature;
- magnetic field (previously was read only from MDS);
- atom/ion fluxes (B2), neutral atom densities (Eirene, if available),
- neutral atom fluxes (Eirene, if available);
- neutral atom temperatures (Eirene, if available).

_The fluxes are not stored in `SOLPSSimulation`, but used to calculate velocities._

Issue #36 is fixed. The parallel velocity and the B2 fluxes (in s<sup>-1</sup>) defined on cell faces are used in `b2_flux_to_velocity()` to calculate velocities at cell centres. If neutral atom fluxes from Eirene (in m<sup>-2</sup> s<sup>-1</sup>) defined at cell centres are available, they are used in `eirene_flux_to_velocity()` replacing the B2 ones.

Issue #37 is fixed. Now all arrays are row-major. However, the indexing is inverted. The correct index order now is: (radial, poloidal) or (species, radial, poloidal) for scalars and (vector component, radial, poloidal) or (species, vector component, radial, poloidal) for vectors. This should improve the performance in typical workflows.

**Remaining issues:**

- Electron velocities are still not read from SOLPS data.

- Molecular and total H-alpha emission are still not read from SOLPS data.


vsnever:
Just updated the code by adding ion and neutral atom temperatures. The latter is not available in stand-alone B2.5 simulations.

vsnever:
I renamed this PR because now it addresses not just #31 but in some degree #8 too.

Replacing the algorithm for poloidal basis vectors calculation in `SOLPSMesh` was more a bug fix than an improvement. The SOLPS mesh have break points due to the separatrix. In the old algorithm, which calculated the basis vectors using the cell centres, the basis vectors had wrong values at these break points.

vsnever:
Just noticed that conversion from poloidal to Cartesian coordinates in `mdsplus.py` is incorrect. It's for orthogonal basis, but SOLPS uses curvilinear coordinates.

vsnever:
> Just noticed that conversion from poloidal to Cartesian coordinates in `mdsplus.py` is incorrect. It's for orthogonal basis, but SOLPS uses curvilinear coordinates.

Sorry, I'm wrong here. It converts **to** Cartesian, so it's OK.

mattngc:
Hey Vlad, I don't have any strong opinions on these improvements. I don't use this module as much at the moment, so I think opinions from the main users are more important. I'll let @jacklovell and @Mateasek give feedback on these changes.

vsnever:
> Hey Vlad, I don't have any strong opinions on these improvements. I don't use this module as much at the moment, so I think opinions from the main users are more important. I'll let @jacklovell and @Mateasek give feedback on these changes.

Hi Matthew, thanks for reply. I need a couple more days to get this PR ready for review.

vsnever:
In the first post, I summaraised the changes made in this PR and marked it ready for review.

Hi @jacklovell and @Mateasek, I am awaiting your verdict).

Hello @jrh-ccfe, could you please share any balance.nc file, so I could improve `load_solps_from_balance()`. I have access to Heimdall cluster, so a path to the file will be enough.

cherab/solps/eirene/parser/fort44.py
```diff
@@ -57,13 +57,17 @@ def assign_fort44_parser(file_version):
     :param file_version: Fort44 file version from the file header.
     :return: Parsing function object
     """
+    fort44_supported_versios = (
```
CnlPepper: typo in name: "versios" -> "versions"

cherab/solps/eirene/parser/fort44.py
```diff
@@ -57,13 +57,17 @@ def assign_fort44_parser(file_version):
...
+        20081111,
+        20130210,
+        20170328,
+    )
 
-    fort44_parser_library = {
-        20170328: load_fort44_2017,
-        20130210: load_fort44_2013
-    }
+    if file_version not in fort44_supported_versios:
```
CnlPepper: versios -> versions

cherab/solps/formats/balance.py
```diff
@@ -93,65 +61,111 @@ def load_solps_from_balance(balance_filename):
...
+    sim = SOLPSSimulation(mesh, species_list)
+
+    # TODO: add code to load SOLPS velocities and magnetic field from files
 
-    # Make Mesh Interpolator function for inside/outside mesh test.
-    inside_outside_data = np.ones(mesh.num_tris)
-    inside_outside = AxisymmetricMapper(Discrete2DMesh(mesh.vertex_coords, mesh.triangles, inside_outside_data, limit=False))
-    sim._inside_mesh = inside_outside
+    # Load electron species
+    sim.set_electron_temperature(fhandle.variables['te'].data.copy() / el_charge)
```
CnlPepper: shouldn't this be a property not a setter?

```
sim.electron_temperature = ....
```

CnlPepper: ignore me it this needs to be optimised for speed.... but I suspect it doesn't need to be.

cherab/solps/solps_plasma.py
```diff
@@ -32,139 +31,219 @@
...
+
+        return self._ion_temperature
+
+    def set_ion_temperature(self, value):
+        _check_array("ion_temperature", value, (self.mesh.nx, self.mesh.ny))
+
+        self._ion_temperature = value
 
     @property
-    def radial_particle_flux(self):
```
CnlPepper, line 122: Setter and getters are not python style. These really should be properties:

```
@property
def myvar(self):
    ......

@myvar.setter
def myvar(self, value):
    ......
```

cherab/solps/eirene/parser/fort44.py
vsnever: Thanks, this is corrected.

cherab/solps/formats/balance.py
vsnever: There is a discussion in #20 and also in comments to the `SOLPSSimulation` code, that `sim.electron_temperature`, etc. should return not the data arrays but the interpolators. However, this is not decided yet. If `sim.electron_temperature` return the interpolator, than setting the electron temperature data in the following style:
```
sim.electron_temperature = data
```
will lead to magic hat paradox, because the user puts one object there, but then gets the different object.

vsnever: Probably the best thing would be to have different properties for data and interpolators, like
```
sim.electron_temperature_data  # for grid data
sim.electron_temperature  # for interpolator
```
@jacklovell, @Mateasek, what do you think?

CnlPepper: I would implement it as you suggest in the most recent comment. If users wish to obtain either an interpolator or the raw data, they should be accessible via separate attributes.

I am curious why these values are not being set via the ```__init__``` rather than updating each attribute. Is there a plan for people to write data by creating an empty object first? If not it may be cleaner to configure the object via an init - like the Equilibrium class. i.e. raw data passed to init.

CnlPepper: Actually ignore the last point, it is just style and we use a similar pattern for the Plasma object.

CnlPepper: Ok I've just had a better look through the code to remind myself. I think there is some confusion here. The simulation object was just meant to be a thin wrapper holding the raw data. So all the properties should be data. The data gets converted to functions when you ask for a plasma object. @Mateasek was asking for additional methods / properties that return the data in functional form i.e. splitting out the code from the create_plasma() method. Can I suggest the following:

```
sim.electron_temperature             # property, raw data arrays
sim.electron_temperature_as_f2d()    # method, function 2d
sim.electron_temperature_as_f3d()    # method, function 3d
```

vsnever: I like this interface. I will replace the `set _... ()` methods with property setters and remove the `get _... ()` methods. I suggest a separate PR for adding the interpolators, because we have to decide whether we use `Discrete2DMesh` and `AxisymmetricMapper` or `SOLPSFunction3D` (see my comment in #20).

CnlPepper: I would use the ```SOLPSFunction3D``` for these. 

Btw I've always been very uncomfortable with the use of the ```AxisymmetricMapper``` and ```Discrete2DMesh``` functions in ```SOLPSFunction*```. Using a float to hold an integer is dangerous, it will (with absolute certainty) result in the wrong data being indexed. I would rewrite the ```SOLPSFunction*``` classes. Have a quick look at the ```Discrete2DMesh``` code:
https://github.com/raysect/source/blob/development/raysect/core/math/function/float/function2d/interpolate/discrete2dmesh.pyx

It is trivial to get it to just return the index. See line 174. I would use the KDtree directly in the SOLPS function, copying the relevant code from``` Discrete2DMesh```:

https://github.com/raysect/source/blob/development/raysect/core/math/function/float/function2d/interpolate/common.pyx

Make the SOLPS functions 2D and them use the axisymmetric wrapper around the SOLPSFunction2D to convert it to 3D, where the result is actually a double!

vsnever: I implemented `SOLPSFunction2D` and `SOLPSVectorFunction2D` classes. Also added `*_f2d()` and `*_f3d()` interpolators for plasma parameters. The demos work, but more testing will probably be needed, which I'll leave for tomorrow.

CnlPepper: Nice work. I see you went with properties due to the low overhead of instancing the existing inside function. All looks good, much cleaner and more reliable.

vsnever:
I updated balance.py and revert the change in the fort44 parser.

cherab/solps/b2/parse_b2_block_file.py
```diff
@@ -47,16 +47,16 @@ def _make_solps_data_object(_data):
 
         # Multiple 2D data field (e.g. na)
         if number > nxyg:
-            _data = np.array(_data).reshape((nxg, nyg, int(number / nxyg)), order='F')
+            _data = np.array(_data).reshape((int(number / nxyg), nyg, nxg))
```
jacklovell: I would use floor division here, instead of manually doing float division and truncating to int. It'll produce cleaner code: compare `int(number / nxyg)` with `number // nxyg`.

cherab/solps/b2/parse_b2_block_file.py
```diff
@@ -47,16 +47,16 @@ def _make_solps_data_object(_data):
...
             if debug:
-                print('Mesh data field {} with dimensions:  {:d} x {:d} x {:d}'.format(name, nxg, nyg, int(number/nxyg)))
+                print('Mesh data field {} with dimensions:  {:d} x {:d} x {:d}'.format(name, int(number / nxyg), nyg, nxg))
```
jacklovell: Same comment re: floor division.

cherab/solps/eirene/parser/fort44_2013.py
```diff
@@ -16,6 +16,7 @@
 # See the Licence for the specific language governing permissions and limitations
 # under the Licence.
 
+import numpy as np
```
jacklovell: This numpy import is not needed.

cherab/solps/eirene/parser/fort44_2017.py
```diff
@@ -16,6 +16,7 @@
 # See the Licence for the specific language governing permissions and limitations
 # under the Licence.
 
+import numpy as np
```
jacklovell: Also not needed.

cherab/solps/eirene/eirene.py
```diff
@@ -91,7 +91,7 @@ def __init__(self, nx, ny, na, nm, ni, ns, species_labels, version=None,
     @property
```
jacklovell, line 89: There's also an unused numpy import in this file.

cherab/solps/formats/balance.py
```diff
@@ -19,139 +19,185 @@
 
 import numpy as np
 from scipy.io import netcdf
-from scipy.constants import elementary_charge
+from scipy import constants
 from raysect.core.math.function.float import Discrete2DMesh
 
 from cherab.core.math.mappers import AxisymmetricMapper
```
jacklovell: These two imports are no longer needed.

cherab/solps/formats/balance.py
```diff
@@ -19,139 +19,185 @@
...
+
+        species_list = []
+        neutral_indx = []
+        am = np.round(fhandle.variables['am'].data).astype(np.int)  # Atomic mass number
+        charge = fhandle.variables['za'].data.astype(np.int)   # Ionisation/charge
+        species_names = fhandle.variables['species'].data.copy()
+        ns = am.size
+        for i in range(ns):
+            symbol = ''.join([b.decode('utf-8').strip(' 0123456789+-') for b in species_names[i]])  # also strips isotope number
+            if symbol != 'D' and symbol != 'T':
```
jacklovell: More Pythonic and extensible: `if symbol not in ('D', 'T')`

cherab/solps/formats/balance.py
```diff
@@ -19,139 +19,185 @@
...
+                isotope = lookup_isotope(symbol, number=am[i])  # will through an error for D or T
```
jacklovell: Typo: throw

cherab/solps/formats/balance.py
```diff
@@ -19,139 +19,185 @@
...
+            if 'b2stel_she_bal' in fhandle.variables:
+                b2_ploss = np.sum(fhandle.variables['b2stel_she_bal'].data, axis=0) / mesh.vol
+
+            if 'b2stel_sna_ion_bal' in fhandle.variables:
+                potential_loss = np.sum(fhandle.variables['b2stel_sna_ion_bal'].data, axis=0) / mesh.vol
+
+            # Save total radiated power to the simulation object
+            total_rad = rydberg_energy * el_charge * potential_loss - b2_ploss
+
+    if isinstance(total_rad, np.ndarray):
```
jacklovell: Setting sim.total_radiation will convert to an ndarray if it isn't already. I'm not sure what else this is checking. And the indentation means the line will also be called when `b2_standalone` is `True`, which will result in an `UnboundLocalError` as `total_rad` isn't even assigned in that case. So the indentation needs to be increased by 8 (I think).

cherab/solps/formats/mdsplus.py
```diff
@@ -17,20 +17,16 @@
 # See the Licence for the specific language governing permissions and limitations
 # under the Licence.
 
-import re
 import numpy as np
-from math import sqrt
 from raysect.core import Point2D
```
jacklovell: Point2D is no longer used.

cherab/solps/formats/mdsplus.py
```diff
@@ -40,215 +36,173 @@ def load_solps_from_mdsplus(mds_server, ref_number):
     :rtype: SOLPSSimulation
     """
 
-    from MDSplus import Connection as MDSConnection
+    from MDSplus import Connection as MDSConnection, mdsExceptions
```
jacklovell: `mdsExceptions` is only available in MDSplus 7, but at Culham we use MDSplus 6.1. You could use `MDSplus.MdsException` instead: it's compatible with both 6.1 and 7 and is a superclass of `TreeNNF`.

cherab/solps/formats/mdsplus.py
```diff
@@ -40,215 +36,173 @@ def load_solps_from_mdsplus(mds_server, ref_number):
...
 
     # Load SOLPS mesh geometry and lookup arrays
-    mesh = load_mesh_from_mdsplus(conn)
-    sim = SOLPSSimulation(mesh)
-    ni = mesh.nx
-    nj = mesh.ny
+    mesh = load_mesh_from_mdsplus(conn, mdsExceptions)
+
+    # Load each plasma species in simulation
+    ns = conn.get('\SOLPS::TOP.IDENT.NS').data()  # Number of species
```
jacklovell: While we're doing a tidy-up, these should be r-strings instead of standard strings. Otherwise Python tries to treat the backslash as an escape character. For robustness it should be:
`ns = conn.get(r'\SOLPS::TOP.IDENT.NS').data()`.

This is true for all the `conn.get` calls.

cherab/solps/formats/mdsplus.py
```diff
@@ -40,215 +36,173 @@ def load_solps_from_mdsplus(mds_server, ref_number):
...
-    sim._electron_density = np.swapaxes(conn.get('\SOLPS::TOP.SNAPSHOT.NE').data(), 0, 1)  # (32, 98) => (98, 32)
-
-    ############################
-    # Load each plasma species #
-    ############################
-
-    # Master list of species, e.g. ['D0', 'D+1', 'C0', 'C+1', ...
-    species_list = conn.get('\SOLPS::TOP.IDENT.SPECIES').data()
+    sim.b_field = conn.get('\SOLPS::TOP.SNAPSHOT.B').data()[:3]
+    # sim.b_field_cylindrical is created authomatically
```
jacklovell: Typo: automatically.

cherab/solps/formats/mdsplus.py
```diff
@@ -40,215 +36,173 @@ def load_solps_from_mdsplus(mds_server, ref_number):
...
+
+    if radial_flux.shape[1] == ny - 1:
+        radial_flux = np.concatenate((np.zeros((ns, 1, nx)), radial_flux), axis=1)
+
+    # Setting velocities from B2 flux
+    sim.b2_flux_to_velocity(poloidal_flux, radial_flux, parallel_velocity)
+
+    # Obtaining additional data from EIRENE and replacing data for neutrals
+
+    b2_standalone = False
```
jacklovell: I'd prefer this to become a `try/except/else` block, with `b2_standalone = False` set in the `else` section. It reads a bit more naturally this way: "try: reading Eirene data. exception: b2 standalone. No exception: not b2 stadalone".

cherab/solps/formats/mdsplus.py
```diff
@@ -17,20 +17,16 @@
...
 from raysect.core import Point2D
-from raysect.core.math.function.float import Discrete2DMesh
 
-from cherab.core.math.mappers import AxisymmetricMapper
+from cherab.core.atomic.elements import lookup_isotope
 from cherab.solps.mesh_geometry import SOLPSMesh
-from cherab.solps.solps_plasma import SOLPSSimulation
+from cherab.solps.solps_plasma import SOLPSSimulation, prefer_element
 
+from matplotlib import pyplot as plt
```
jacklovell: pyplot is unused.

cherab/solps/formats/raw_simulation_files.py
```diff
@@ -19,33 +19,16 @@
 
 import os
 import numpy as np
+from scipy.constants import elementary_charge
 from raysect.core.math.function.float import Discrete2DMesh
 
 from cherab.core.math.mappers import AxisymmetricMapper
```
jacklovell: These imports are no longer used.

cherab/solps/formats/raw_simulation_files.py
```diff
@@ -56,93 +39,154 @@ def load_solps_from_raw_output(simulation_path, debug=False):
...
-    * Eirene output file (fort.44)
+    * Eirene output file (fort.44), optional
 
     :param str simulation_path: String path to simulation directory.
     :rtype: SOLPSSimulation
     """
 
     if not os.path.isdir(simulation_path):
-        RuntimeError("simulation_path must be a valid directory")
+        RuntimeError("Simulation_path must be a valid directory.")
```
jacklovell: `simulation_path` is the (case sensitive) name of the argument to this function. It should be left as is. Agree with the additional full stop though.

jacklovell: Also, there's a missing `raise` statement here.

cherab/solps/formats/raw_simulation_files.py
```diff
@@ -56,93 +39,154 @@ def load_solps_from_raw_output(simulation_path, debug=False):
...
     b2_state_file = os.path.join(simulation_path, 'b2fstate')
     eirene_fort44_file = os.path.join(simulation_path, "fort.44")
 
     if not os.path.isfile(mesh_file_path):
-        raise RuntimeError("No B2 b2fgmtry file found in SOLPS output directory")
+        raise RuntimeError("No B2 b2fgmtry file found in SOLPS output directory.")
 
     if not(os.path.isfile(b2_state_file)):
-        RuntimeError("No B2 b2fstate file found in SOLPS output directory")
+        RuntimeError("No B2 b2fstate file found in SOLPS output directory.")
```
jacklovell: Missing `raise`.

cherab/solps/formats/raw_simulation_files.py
```diff
@@ -56,93 +39,154 @@ def load_solps_from_raw_output(simulation_path, debug=False):
...
 
     mesh_file_path = os.path.join(simulation_path, 'b2fgmtry')
     b2_state_file = os.path.join(simulation_path, 'b2fstate')
     eirene_fort44_file = os.path.join(simulation_path, "fort.44")
 
     if not os.path.isfile(mesh_file_path):
-        raise RuntimeError("No B2 b2fgmtry file found in SOLPS output directory")
+        raise RuntimeError("No B2 b2fgmtry file found in SOLPS output directory.")
 
     if not(os.path.isfile(b2_state_file)):
```
jacklovell: Superfluous parentheses.

cherab/solps/formats/raw_simulation_files.py
```diff
@@ -56,93 +39,154 @@ def load_solps_from_raw_output(simulation_path, debug=False):
...
-        RuntimeError("No B2 b2fstate file found in SOLPS output directory")
+        RuntimeError("No B2 b2fstate file found in SOLPS output directory.")
 
     if not(os.path.isfile(eirene_fort44_file)):
```
jacklovell: Superfluous parentheses.

cherab/solps/mesh_geometry.py
```diff
@@ -18,16 +18,12 @@
 # under the Licence.
 
 # External imports
-from collections import namedtuple
+# from collections import namedtuple
```
jacklovell: Delete rather than comment out.

cherab/solps/mesh_geometry.py
```diff
@@ -40,89 +36,122 @@ class SOLPSMesh:
...
+                             must be 3 dimensional. Example shape is (4 x 32 x 98).
     """
 
-    def __init__(self, cr_r, cr_z, vol):
+    # TODO Make neighbix and neighbix optional in the future, as they can be reconstructed with _tri_index_loopup
+
+    def __init__(self, r, z, vol, neighbix, neighbiy):
+
+        if r.shape != z.shape:
+            raise ValueError('Shape of r array: %s mismatch the shape of z array: %s.' % (r.shape, z.shape))
```
jacklovell: Please use string.format rather than %-style formatting, for consistency with the rest of the Cherab code base.

cherab/solps/mesh_geometry.py
```diff
@@ -40,89 +36,122 @@ class SOLPSMesh:
...
+        vec_z = z[2] - z[0]
+        vec_magn = np.sqrt(vec_r**2 + vec_z**2)
+        self._radial_basis_vector[0] = np.divide(vec_r, vec_magn, out=np.zeros_like(vec_magn), where=(vec_magn > 0))
+        self._radial_basis_vector[1] = np.divide(vec_z, vec_magn, out=np.zeros_like(vec_magn), where=(vec_magn > 0))
+
+        # Calculating poloidal contact areas
+        self._poloidal_area = np.pi * (r[2] + r[0]) * vec_magn
+
+        # For convertion from Cartesian to poloidal
+        # TODO Make it work with trianle cells
```
jacklovell: Typo: triangle.

cherab/solps/mesh_geometry.py
```diff
@@ -40,89 +36,122 @@ class SOLPSMesh:
...
+        self._inv_det = 1. / (self._poloidal_basis_vector[0] * self._radial_basis_vector[1] -
+                              self._poloidal_basis_vector[1] * self._radial_basis_vector[0])
+
+        # Test for basis vector calculation
+        # plt.quiver(self._cr[0], self._cz[0], self._radial_basis_vector[0, 0], self._radial_basis_vector[1, 0], color='k')
+        # plt.quiver(self._cr[0], self._cz[0], self._poloidal_basis_vector[0, 0], self._poloidal_basis_vector[1, 0], color='r')
+        # plt.quiver(self._cr[-1], self._cz[-1], self._radial_basis_vector[0, -1], self._radial_basis_vector[1, -1], color='k')
+        # plt.quiver(self._cr[-1], self._cz[-1], self._poloidal_basis_vector[0, -1], self._poloidal_basis_vector[1, -1], color='r')
+        # plt.gca().set_aspect('equal')
+        # plt.show()
```
jacklovell: Delete unused test code.

cherab/solps/mesh_geometry.py
```diff
@@ -40,89 +36,122 @@ class SOLPSMesh:
...
+    :param ndarray neighbix: Array of poloidal indeces of neighbouring cells in order: left, bottom, right, top,
+                             must be 3 dimensional. Example shape is (4 x 32 x 98).
+                             In SOLPS notation: left/right - poloidal prev./next, bottom/top - radial prev./next.
+                             Cell indexing starts with 0 and -1 means no neighbour.
+    :param ndarray neighbix: Array of radial indeces of neighbouring cells in order: left, bottom, right, top,
+                             must be 3 dimensional. Example shape is (4 x 32 x 98).
     """
 
-    def __init__(self, cr_r, cr_z, vol):
+    # TODO Make neighbix and neighbix optional in the future, as they can be reconstructed with _tri_index_loopup
```
jacklovell: `_tri_index_lookup` no longer exists. Perhaps modify this comment to `_triangle_to_grid_map`.

cherab/solps/mesh_geometry.py
```diff
@@ -135,70 +164,161 @@ def cz(self):
...
         # Code for plotting vessel geometry if available
-        # for i in range(vessel.shape[0]):
-        #     plt.plot([vessel[i, 0], vessel[i, 2]], [vessel[i, 1], vessel[i, 3]], 'k')
-        # for i in range(vessel.shape[0]):
-        #     plt.plot([vessel[i, 0], vessel[i, 2]], [vessel[i, 1], vessel[i, 3]], 'or')
+        # if self.vessel is not None:
+        #     for i in range(self.vessel.shape[0]):
+        #         ax.plot([self.vessel[i, 0], self.vessel[i, 2]], [self.vessel[i, 1], self.vessel[i, 3]], 'k')
+        #     for i in range(self.vessel.shape[0]):
+        #         ax.plot([self.vessel[i, 0], self.vessel[i, 2]], [self.vessel[i, 1], self.vessel[i, 3]], 'or')
```
jacklovell: Delete commented out code, or un-comment it.

cherab/solps/solps_2d_functions.pyx
```diff
@@ -0,0 +1,214 @@
...
+    @cython.initializedcheck(False)
+    cdef double evaluate(self, double x, double y) except? -1e999:
+
+        cdef:
+            np.int32_t triangle_id, ix, iy
+
+        if self._kdtree.is_contained(new_point2d(x, y)):
+
+            triangle_id = self._kdtree.triangle_id
+            ix, iy = self._triangle_to_grid_map_mv[triangle_id, :]
```
jacklovell: Tuple unpacking is unfortunately not optimised by Cython: this still goes through a high level Python object (see the HTML output of `cython -a solps_2d_functions.pyx`. A more efficient solution would be:
```python
indices = self._triangle_to_grid_map_mv[triangle_id, :]
ix = indices[0]
iy = indices[1]
```
Of course, `indices` would need to be `cdef`'d appropriately.

cherab/solps/solps_2d_functions.pyx
```diff
@@ -0,0 +1,214 @@
...
+    cdef Vector3D evaluate(self, double x, double y):
+
+        cdef:
+            np.int32_t triangle_id, ix, iy
+            double vx, vy, vz
+
+        if self._kdtree.is_contained(new_point2d(x, y)):
+
+            triangle_id = self._kdtree.triangle_id
+            ix, iy = self._triangle_to_grid_map_mv[triangle_id, :]
```
jacklovell: See previous comment on tuple unpacking.

cherab/solps/solps_3d_functions.pyx
```diff
@@ -101,7 +101,9 @@ cdef class SOLPSVectorFunction3D(VectorFunction3D):
 
         # print(self._grid_vectors.shape)
         # Lookup vector for this grid cell.
-        vx, vy, vz = self._grid_vectors[ix, iy, :]
+        vx = self._grid_vectors[0, ix, iy]
+        vy = self._grid_vectors[1, ix, iy]
+        vz = self._grid_vectors[2, ix, iy]
```
jacklovell: Do the same unrolling on the `ix, iy` assignments in both `evaluate` methods.

cherab/solps/solps_plasma.py
```diff
@@ -17,228 +17,577 @@
 # See the Licence for the specific language governing permissions and limitations
 # under the Licence.
 
-import re
 import pickle
 import numpy as np
 import matplotlib.pyplot as plt
```
jacklovell: pyplot is unused.

cherab/solps/solps_plasma.py
```diff
@@ -17,228 +17,577 @@
...
-from scipy.constants import atomic_mass, electron_mass
+from scipy.constants import atomic_mass, electron_mass, elementary_charge
```
jacklovell: Elementary charge is unused: left over from development?

cherab/solps/solps_plasma.py
```diff
@@ -17,228 +17,577 @@
...
 
 # Raysect imports
-from raysect.core.math.function.float import Discrete2DMesh
 from raysect.core import translate, Point3D, Vector3D, Node, AffineMatrix3D
 from raysect.primitive import Cylinder
 from raysect.optical import Spectrum
```
jacklovell: Of these, only `translate`, `Vector3D` and `Cylinder` are actually used. The rest can be removed.

cherab/solps/solps_plasma.py
```diff
@@ -17,228 +17,577 @@
...
+        self._total_radiation = None
+        self._total_radiation_f2d = None
+        self._total_radiation_f3d = None
+        self._b_field = None
+        self._b_field_cylindrical = None
+        self._b_field_cylindrical_f2d = None
+        self._b_field_cartesian = None
+        self._eirene_model = None  # what is this for?
+        self._b2_model = None  # what is this for?
+        self._eirene = None  # do we need this in SOLPSSimulation?
```
jacklovell: I think it's fine to leave these here: they provide slightly more direct access to the underlying simulation data and can be used for debugging purposes.

cherab/solps/solps_plasma.py
```diff
@@ -17,228 +17,577 @@
...
     @property
     def total_radiation(self):
         """
-        Total radiation
+        Total radiation at each mesh cell.
+        Array of shape (ny, nx).
 
         This is not calculated from the CHERAB emission models, instead it comes from the SOLPS output data.
         Is calculated from the sum of all integrated line emission and all Bremmstrahlung. The signals used are 'RQRAD'
         and 'RQBRM'. Final output is in W/str?
```
jacklovell: I think this should be in W/m3, based on #28.

cherab/solps/solps_plasma.py
```diff
@@ -17,228 +17,577 @@
...
         """
-        if self._total_rad is None:
-            raise RuntimeError("Total radiation not available for this simulation")
+        if self._total_radiation is None:
+            raise RuntimeError("Total radiation not available for this simulation.")
         else:
```
jacklovell: `else` is not necessary after a `raise`. This applies to all the `total_radiation*`, `b_field*` and `eirene_simulation` properties.

cherab/solps/solps_plasma.py
```diff
@@ -248,55 +597,201 @@ def eirene_simulation(self):
...
+        bottomix = mesh.neighbix[1]  # radial prev.
+        bottomiy = mesh.neighbiy[1]
+        rightix = mesh.neighbix[2]   # poloidal next.
+        rightiy = mesh.neighbiy[2]
+        topix = mesh.neighbix[3]  # radial next.
+        topiy = mesh.neighbiy[3]
+
+        # Converting s-1 --> m-2 s-1
+        poloidal_flux = np.divide(poloidal_flux, poloidal_area, out=np.zeros_like(poloidal_flux), where=poloidal_area > 0)
+        radial_flux = np.divide(radial_flux, radial_area, out=np.zeros_like(radial_flux), where=radial_area > 0)
```
jacklovell: Does specifying the `out` argument here actually help? You're still having to allocate a new array, and I thought the `out` argument was only relevant if there was an existing array to put the result in to avoid a new allocation. There are a few other places below where you also do this.

cherab/solps/solps_plasma.py
```diff
@@ -460,77 +955,66 @@ def create_plasma(self, parent=None, transform=None, name=None):
...
-        plasma.electron_distribution = Maxwellian(electron_dens, electron_temp, electron_velocity, electron_mass)
-
-        if not isinstance(self.velocities_cartesian, np.ndarray):
-            print('Warning! No velocity field data available for this simulation.')
-
-        b2_neutral_i = 0  # counter for B2 neutrals
-        for k, sp in enumerate(self.species_list):
+        if self.electron_velocities_cartesian is None:
+            print('Warning! No electron velocity data available for this simulation.')
+            electron_velocity = lambda x, y, z: Vector3D(0, 0, 0)
```
jacklovell: This should be a `ConstantVector3D` object, not a lambda function.

cherab/solps/solps_plasma.py
```diff
@@ -460,77 +955,66 @@ def create_plasma(self, parent=None, transform=None, name=None):
...
-            # dens = SOLPSFunction3D(tri_index_lookup, tri_to_grid, species_dens_data)
+            charge = sp[1]
 
             # Create the velocity vector lookup function
-            if isinstance(self.velocities_cartesian, np.ndarray):
-                velocity = SOLPSVectorFunction3D(tri_index_lookup, tri_to_grid, self.velocities_cartesian[:, :, k, :])
+            if self.velocities_cartesian is not None:
+                velocity = self.velocities_cartesian[k]
             else:
                 velocity = lambda x, y, z: Vector3D(0, 0, 0)
```
jacklovell: This should be a `ConstantVector3D` object, not a lambda function.

cherab/solps/solps_plasma.py
```diff
@@ -460,77 +955,66 @@ def create_plasma(self, parent=None, transform=None, name=None):
...
-
-        b2_neutral_i = 0  # counter for B2 neutrals
-        for k, sp in enumerate(self.species_list):
+        if self.electron_velocities_cartesian is None:
+            print('Warning! No electron velocity data available for this simulation.')
+            electron_velocity = lambda x, y, z: Vector3D(0, 0, 0)
+        else:
+            electron_velocity = self.electron_velocities_cartesian
+        plasma.electron_distribution = Maxwellian(self.electron_density_f3d, self.electron_temperature_f3d, electron_velocity, electron_mass)
+
```
jacklovell: Trailing whitespace.

jacklovell (changes requested):
Thanks for this Vlad: it's a huge amount of great work and a big improvement to the module overall. I've made some in-line comments on the code, but generally it is of a very good standard.

I've run the 3 demo scripts we currently have on a limited set of runs that I have access to and they seem to produce mostly the same results as before. I expect any small differences are a result of bug fixes. Have you done regression testing on any other SOLPS cases you have?

There's a few places in the code with large commented-out blocks: these can be removed if they won't be added back in later. I'm thinking in particular in solps_plasma.py, where there are several commented-out plotting methods which should live in demo scripts instead.

Also, considering the amount of change here it would be worth adding to the changelog while the changes are still fresh. I'll defer to @CnlPepper on whether this is a big enough change to warrant a 1.2 (or even a 2.0) release, or whether a 1.1.1 would be suitable.

jacklovell:
One other comment: if Axisymmetric wrappers of  `SOLPSFunction2D` and `SOLPSVectorFunction2D` replace the corresponding 3D functions, perhaps we should remove the latter from the module entirely. The 3D functions are still present in this PR.

CnlPepper:
> One other comment: if Axisymmetric wrappers of `SOLPSFunction2D` and `SOLPSVectorFunction2D` replace the corresponding 3D functions, perhaps we should remove the latter from the module entirely. The 3D functions are still present in this PR.

I agree, it would be best to remove redundant code.

CnlPepper:
> Also, considering the amount of change here it would be worth adding to the changelog while the changes are still fresh. I'll defer to @CnlPepper on whether this is a big enough change to warrant a 1.2 (or even a 2.0) release, or whether a 1.1.1 would be suitable.

While there are large API changes to the SOLPSSimulation API, the core functionality for loading and generating Plasma objects 
 and the object layout are essentially the same. It is definitely a new minor release, possibly a major release.

The arguments for a major/minor version increment would be the degree of impact to the main usage of the package. This one lies on the border I suspect. As the active users of this package, you would be best placed to assess the impact on users and then decide if you want to go for v2.0.0 or v1.2.0.

Please do add a changelog. I would recommend getting into the habit of checking if it needs updating with every pull request.

cherab/solps/mesh_geometry.py
```diff
@@ -40,89 +36,122 @@ class SOLPSMesh:
...
-    :param ndarray cr_z: Array of cell vertex z coordinates, must be 3 dimensional. Example shape is (98 x 32 x 4).
-    :param ndarray vol: Array of cell volumes. Example shape is (98 x 32).
+    :param ndarray r: Array of cell vertex r coordinates, must be 3 dimensional. Example shape is (4 x 32 x 98).
+    :param ndarray z: Array of cell vertex z coordinates, must be 3 dimensional. Example shape is (4 x 32 x 98).
+    :param ndarray vol: Array of cell volumes. Example shape is (32 x 98).
+    :param ndarray neighbix: Array of poloidal indeces of neighbouring cells in order: left, bottom, right, top,
+                             must be 3 dimensional. Example shape is (4 x 32 x 98).
+                             In SOLPS notation: left/right - poloidal prev./next, bottom/top - radial prev./next.
+                             Cell indexing starts with 0 and -1 means no neighbour.
+    :param ndarray neighbix: Array of radial indeces of neighbouring cells in order: left, bottom, right, top,
```
Mateasek: shouldn't it be neighbiy?

cherab/solps/solps_plasma.py
```diff
@@ -248,55 +597,201 @@ def eirene_simulation(self):
...
             return self._eirene
 
+    @eirene_simulation.setter
+    def eirene_simulation(self, value):
+        if not isinstance(value, Eirene):
+            raise ValueError('Attribute "eirene_simulation" must be an Eirene instance.')
+
+        self._eirene = value
+
+    def b2_flux_to_velocity(self, poloidal_flux, radial_flux, parallel_velocity):
```
Mateasek: Wouldn't it be better to make this method a standalone instead of class method, like it is in the case of EFITEquilibrium class? This class is becoming quite big, it would help decreasing it. Also the method itself could be useful/used for various purposes. The same could be done with `eirene_flux_to_velocity`.

Mateasek:
That is a large amount of work @vsnever, very nice. Thanks for all the work you put into it, especially when SOLPS is not the easiest code to understand and read! I would like to propose one more change, since this is quite major rewrite. I would suggest also removing the commented plotting methods from the SOLPSimulation class. I think the plots would be better done outside in a separate file, as it is with equilibrium plots in Cherab. I don't mean that you should be also adding it now.

cherab/solps/formats/balance.py
vsnever: `total_rad` is calculated even if `b2_standalone` is `True` (without neutrals), but the indentation was wrong indeed, this must be inside `with` statement. Corrected.

cherab/solps/solps_plasma.py
vsnever: I corrected this to W m-3, however this docstring is valid only if `total_radiation` is read from MDSPlus server. When obtained from raw output, `total_radiation` includes only radiation from neutral atoms, when obtained from balance.nc, it does not include bremsstrahlung.

cherab/solps/solps_plasma.py
vsnever: This code was before me, some properties threw errors when not initialized and some did not. I think, they should not throw errors. If a property is required for a function to work, that function must check if the value is set and throw a `RuntimeError` if not. I've removed these `raise` statements from properties and instead added them to functions that don't work without these properties.

cherab/solps/solps_plasma.py
vsnever: No, if I do not specify the `out` as a zero array explicitly, I'll get uninititalised values where the condition for divide is not met. From numpy.divide description: "...Note that if an uninitialized out array is created via the default out=None, locations within it where the condition is False will remain uninitialized.", [https://numpy.org/doc/stable/reference/generated/numpy.divide.html](https://numpy.org/doc/stable/reference/generated/numpy.divide.html).

vsnever:
@jacklovell, @Mateasek, @CnlPepper, thank you very much for the code review. I hope I addressed all the issues you raised. For three of them I left the comments (see above). If I did not leave a comment, it means that the issue is fixed exactly as you suggested.

>I've run the 3 demo scripts we currently have on a limited set of runs that I have access to and they seem to produce mostly the same results as before. I expect any small differences are a result of bug fixes. Have you done regression testing on any other SOLPS cases you have?

When the data is read from the MDSPlus, the only difference should be in species velocities, because of the bug fix. When the data is read from the balance.nc file, then in addition to velocities, the density of impurity neutral atoms is different, because now it's obtained from the Eirene data. If the data is read from raw simulation files then the density of main plasma neutral atoms is also different because previously it was obtained from B2.

>Also, considering the amount of change here it would be worth adding to the changelog while the changes are still fresh. I'll defer to @CnlPepper on whether this is a big enough change to warrant a 1.2 (or even a 2.0) release, or whether a 1.1.1 would be suitable.

I updated the changelog suggesting a 1.2 version.

In addition to the improvements you suggested, I removed redundant check for inside/outside mesh in `SOLPSTotalRadiatedPower` emitter, and because now this emitter uses only a `total_radiation_f3d` from `SOLPSSimulation`, I think that passing the simulation object to this emitter is no longer needed, a Function3D instance is enough.

cherab/solps/solps_plasma.py
jacklovell: Ah, I see. Thanks for clarifying.

jacklovell (approved):
Looks good to me.

jacklovell:
@CnlPepper @Mateasek are you happy for these changes to be merged?

Mateasek (approved):

Mateasek:
Yes, sorry I am new to this, forgot to click on approve. Very nice piece of work!

CnlPepper:
Happy here, nice work!

vsnever:
@jacklovell, @Mateasek, @CnlPepper, thanks a lot!

@jacklovell, please do not close #34 when this is merged. There is still plenty of work to do.

Issue #32, opened 2020-09-03
Make fort44 parser to support more versions of EIRENE fort.44 file


vsnever:
Currently, the parser will throw an error if the version of fort.44 is not equal to 20130210 or 20170328. In b2plot source files I found the references to the following versions of fort.44: 960511, 960513, 960623, 960727, 961228, 20000727, 20051115, 20060206, 20071209, 20080706, 20081111, 20130210, 20170228, 20170328 and 20180323. I doubt it makes sense now to add support for all of these versions, however support can be added upon request. 

I also noticed that instead of requiring an exact version match, b2plot uses `>=` operator. I think the same approach should be implemented in fort44 parser here. For example, I had to manually change the generomak version of the fort.44 file from 20081111 to 20130210 for the fort44 parser to read it. Yet, it was read successfully. I suggest that for now, all files older than 20170328 should be read by `fort44_2013` parser and all other files by `fort44_2017` parser. I think it's better to display a warning if the version of fort.44 does not match any of the versions guaranteed to be supported by the parser than to throw an error.

jacklovell:
We deliberately made the version checks very restrictive, because neither @Mateasek or I could find a comprehensive list of the different fort.44 versions and their specifications. It may be buried in the git history of the b2plot source code, and extracting that wasn't an appealing prospect. We got it to work for the specific versions of fort.44 files we have available and need for our work, and then erred on the side of caution in rejecting all other un-tested versions.

We'd need to be very careful with accepting ranges of file versions: fort.44 is not self-describing and it's possible to read an incompatible format without error which produces a garbage simulation object, on the off chance that the array sizes of data in the file happen to be compatible with whichever fort.44 version we happen to be reading (I've seen this before). We'd have to manually check every data item read from every version to be confident.

The `>=` operator in b2plot does suggest some backward compatibility, so it makes sense to modify the reading logic for this too. But this is inconsistent with our observations and the current code, which suggests that a 2013 file is incompatible with our 2017 reader. It would be good to see how b2plot handles this, and then perhaps we should re-write our readers to behave similarly.

vsnever:
Ok, I'll revert the respective change in #35 (after you review this PR), but at least, let's add 20081111 version in the supported list, because fort44_2013 parser reads it correctly.

Mateasek:
[...]

PR #40, opened 2020-09-30
Fort44 parser for pre 2007 Eirene fort.44 file versions

vsnever:
This adds a fort.44 parser for the Eirene fort.44 file versions older than 20071209. This parser does not read total radiated power by neutral atoms ('eneutrad'), which was introduced in version 20071209. It is suggested to use this parser also for versions 20071209. 20080706 and 20081111 as a workaround until a generic parser is created. See discussion in #32 and #39.

### DIFF
--- a/cherab/solps/eirene/parser/fort44.py
+++ b/cherab/solps/eirene/parser/fort44.py
@@ -18,7 +18,7 @@
 
 from cherab.solps.eirene.parser.fort44_2017 import load_fort44_2017
 from cherab.solps.eirene.parser.fort44_2013 import load_fort44_2013
-from cherab.solps.eirene.parser.fort44_old import load_fort44_old
+from cherab.solps.eirene.parser.fort44_pre_2007 import load_fort44_pre_2007
 
 
 def load_fort44_file(file_path, debug=False):
@@ -59,18 +59,22 @@ def assign_fort44_parser(file_version):
     :return: Parsing function object
     """
 
+    # TODO write universal fort44 parser
+
+    # Parser load_fort44_pre_2007 is used also for 20071209, 20080706 and 20081111
+    # versions as a temporary solution. Parameter 'eneutrad' is not parsed from these file versions.
     fort44_parser_library = {
-        960511: load_fort44_old,
-        960513: load_fort44_old,
-        960623: load_fort44_old,
-        960727: load_fort44_old,
-        961228: load_fort44_old,
-        20000727: load_fort44_old,
-        20051115: load_fort44_old,
-        20060206: load_fort44_old,
-        20071209: load_fort44_old,  # here we are missing eneutrad
-        20080706: load_fort44_old,  # here we are missing eneutrad
-        20081111: load_fort44_old,  # here we are missing eneutrad
+        960511: load_fort44_pre_2007,
+        960513: load_fort44_pre_2007,
+        960623: load_fort44_pre_2007,
+        960727: load_fort44_pre_2007,
+        961228: load_fort44_pre_2007,
+        20000727: load_fort44_pre_2007,
+        20051115: load_fort44_pre_2007,
+        20060206: load_fort44_pre_2007,
+        20071209: load_fort44_pre_2007,
+        20080706: load_fort44_pre_2007,
+        20081111: load_fort44_pre_2007,
         20130210: load_fort44_2013,
         20170328: load_fort44_2017
     }
--- a/cherab/solps/eirene/parser/fort44.py
+++ b/cherab/solps/eirene/parser/fort44.py
@@ -18,6 +18,7 @@
 
 from cherab.solps.eirene.parser.fort44_2017 import load_fort44_2017
 from cherab.solps.eirene.parser.fort44_2013 import load_fort44_2013
+from cherab.solps.eirene.parser.fort44_old import load_fort44_old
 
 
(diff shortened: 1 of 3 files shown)


jacklovell:
This looks OK to me. Which of these versions have you been able to test with existing output files? It would be good to know which ones we've verified and which ones we're assuming will work based on b2plot.

jacklovell:
Also, @Mateasek do you have any thoughts on this PR, or additional file versions to test?

vsnever:
> This looks OK to me. Which of these versions have you been able to test with existing output files? It would be good to know which ones we've verified and which ones we're assuming will work based on b2plot.

I've tested only 20081111, all other versions are assumed to work based on the code from converter (`modules/B2.5/src/convert/b2yt_ngread.F`) and b2plot (`modules/B2.5/src/b2plot/ngread.F`). I don't have files of older versions for testing.

Mateasek:
I can test 20170328 and 20180323 versions. I will try to look at the PR this evening.

jacklovell:
> > This looks OK to me. Which of these versions have you been able to test with existing output files? It would be good to know which ones we've verified and which ones we're assuming will work based on b2plot.
> 
> I've tested only 20081111, all other versions are assumed to work based on the code from converter (`modules/B2.5/src/convert/b2yt_ngread.F`) and b2plot (`modules/B2.5/src/b2plot/ngread.F`). I don't have files of older versions for testing.

OK. I've found the following versions on the UKAEA filesystem which we can also check: 961228, 20000727, 20070912, 20080706. There may be more on other systems.

Also, can you check whether it's 20070912 or 20071209? The PR includes the latter, but the fort.44 files I'm seeing at UKAEA are the former.

vsnever:
> Also, can you check whether it's 20070912 or 20071209? The PR includes the latter, but the fort.44 files I'm seeing at UKAEA are the former.

It's 20071209. If fort.44 version is less than 20071209, `eneutrad` is not read. However, I'm wondering if this is a coincidence or Eirene stamps the version according to the system locale, e.g. YYYY/MM/DD or YYYY/DD/MM?

Mateasek:
I looked into the fort.44 file version 20180323, which already has the headers @jacklovell pointed out. I'm bit behind in the fort.44 knowledge so please correct me if I am wrong or if I am stating something obvious. In the file header it gives (observed from the fort44_2017 file) number of columns/rows 96 and 36 and that there are 2 atoms, 1 molecule and 1 ion in the simulation. All the 18 blocks being read by the parser have in headers 3456 or 6912, which corresponds to 96 * 36 * n, where n is 1 or 2. If n is 1, then I guess the block contains only information connected to a single species or a superposition for more species. Doesn't this mean the files with such headers are self-described as @jacklovell says and we can at least safely parse the file data into data blocks? I know it doesn't say what quantity exactly the numbers give, but this should safely allow us to distinguish if the block resolves more elements/molecules/ions, as @vsnever pointed out.

The Eirene class has now attributes as `edism`, `pem` and etc. I would be up for changing the names to be more descriptive of the quantity contained. I can't really connect `emist` to `# Total Halpha Emission (including molecules)` and cherab users are not necessarily SOLPS users, so any name link is not required imo. Another reason to do this would be to completely disconnect Cherab namespace from the SOLPS namespaces, which to my understanding is quite "evolving" between versions. Then the assignment of the quantities in the Eirene class could be done after the file is parsed and the quantities in the block identified for different versions. What do you think?

Mateasek:
Just to add, what I proposed is kind of what is being done now, but breaking the namespaces more and documenting the mapping somewhere would give us in my opinion a nice opportunity to raise warnings and errors (as @jacklovell suggested) and also to react to differences in data availability between versions when building up plasma object, for example: `Radiated power for individual neutrals is missing, using data from total neutral radiated power`.

vsnever:
> Doesn't this mean the files with such headers are self-described as @jacklovell says and we can at least safely parse the file data into data blocks?

Yes, you are correct. The files with label lines before each data block can be parsed without problem. The only problem is to parse the files with versions in the range from 20071209 to 20160829.

Mateasek:
> The only problem is to parse the files with versions in the range from 20071209 to 20160829.

I have only access to versions 2017+, it will probably be the best to leave up to you two to resolve problems with these versions, I don't see where I could contribute. Anyway, let me know if I could be of any use.

jacklovell:
> > Also, can you check whether it's 20070912 or 20071209? The PR includes the latter, but the fort.44 files I'm seeing at UKAEA are the former.
> 
> It's 20071209. If fort.44 version is less than 20071209, `eneutrad` is not read. However, I'm wondering if this is a coincidence or Eirene stamps the version according to the system locale, e.g. YYYY/MM/DD or YYYY/DD/MM?

Huh. I can see fort.44s with 20070912 and 20071209. No matter: if anybody has a burning desire to read these rogue 20070912 files we can always test and add them later. For now I'll test 961228, 20000727, 20071209, 20080706 and 20081111, as these are all the pre-2013 versions I have access to.

jacklovell:
I've managed to test 961228, 20000727, 20071209 and 20081111: these came from runs on the UKAEA computers which had all 3 required raw files for Cherab to read the simulation. All of these look at least qualitatively sensible to me, so I'm happy to merge the changes. We can implement a more generic reader for labelled fort.44 versions (>=2016) in a separate PR.

PR #41, opened 2020-10-16
Add more general parser for fort.44 files which include labels

jacklovell:
Versions of fort.44 >= 20160829 have a comment line before each data
block which includes the name of the field and the size of the data
block. This makes the file close enough to self-describing that the
block can be parsed without knowledge of the exact format. This should
enable greater flexibility for the number of formats we support.

This change means we essentially have two types of parsers, unlabelled
and labelled. The unlabelled ones are further split into pre-2007 and
2013 at the moment, due to non-self-describing differences in these
formats. But the availability of at least a couple of general parsers
leads to a reworking of the logic to choose a parser: it's no longer
tied to the exact version but is more of a boundary match.

This has been tested on fort.44 files with versions 20160829, 20170328 and 20180323 for SOLPS simulations carried out in DIII-D, ITER and AUG geometries. I also tested it on a WEST simulation which only included He atoms, and it failed loudly because of missing dmb2 etc. entries: I think this is a good thing compared with silently reading garbage. I'm in 2 minds as to whether to add extra complexity to deal with this: should we automatically populate missing fields with zeros, or error out on the basis that these files do not comply with the specification in the SOLPS manual for fort.44? I think it's fair to say we only support compliant files for the purposes of code simplicity.

Here is a [script](https://github.com/cherab/solps/files/5392981/plot_eirene_simulation.py.zip) which can be used for quickly plotting all the quantities we read into the Eirene object, to help with testing other files. It requires the b2fstate, b2fgmtry and fort.44 files to all live in the same directory (so some copying may be required for runs which reference a baserun).

@Mateasek @vsnever could you check that this works with suitable SOLPS outputs you have access to? Other suggestions for improvement are also welcome.


cherab/solps/eirene/parser/fort44_post_2016.py
```diff
@@ -0,0 +1,135 @@
...
+        eirene.rem = raw_data['refluxm'].reshape(eirene.nm, eirene.ny, eirene.nx)
+
+        # Read poloidal energy flux (pea, pem)
+        eirene.pea = raw_data['pefluxa'].reshape(eirene.na, eirene.ny, eirene.nx)
+        eirene.pem = raw_data['pefluxm'].reshape(eirene.nm, eirene.ny, eirene.nx)
+
+        # Halpha total & molecules (emist, emism)
+        # fort.44 gives emission due to atoms and emission due to molecules separately.
+        eirene.emism = raw_data['emissmol'].reshape(1, eirene.ny, eirene.nx)
+        eirene.emist = raw_data['emiss'].reshape(1, eirene.ny, eirene.nx) + eirene.emism
```
vsnever: According to b2plot, 'emiss' is a total H-alpha emissivity. However, I'm not sure if "total" means molecular + atomic or just atomic but summed over all isotopes. When saved to MDSplus format, "\SOLPS::TOP.SNAPSHOT.EMISS" is definitely an atomic emissivity, I've checked that, but this doesn't mean that it's the same for the raw output.

I have to ask Andrey Pshenov or Andrey S. Kukushkin about this.

cherab/solps/eirene/parser/fort44_post_2016.py
```diff
@@ -0,0 +1,135 @@
...
+
+        # Radiated power (elosm, edism, eradt)
+        eirene.edism = raw_data['edissml'].reshape(eirene.nm, eirene.ny, eirene.nx)
+
+        eneutrad = raw_data['eneutrad'].reshape(eirene.na, eirene.ny, eirene.nx)
```
vsnever, line 112: Depending on SOLPS version and Eirene format, `eneutrad` may be resolved or summed over the atoms. So I propose changing this to:
```suggestion
        n_eneutrad = raw_data['eneutrad'].size // (eirene.ny * eirene.nx)
        eneutrad = raw_data['eneutrad'].reshape(n_eneutrad, eirene.ny, eirene.nx)
```

vsnever (commented):
This is a great work. Unfortunately, the only version of fort.44 that I have access to is 20081111, because only two people in Kurchatov who do simulations with SOLPS are using SOLPS 4.3. However, if you can provide me with links to the labelled fort.44 files in Heimdall, I will be happy to do the testing.

vsnever:
> I also tested it on a WEST simulation which only included He atoms, and it failed loudly because of missing dmb2 etc. entries: I think this is a good thing compared with silently reading garbage. I'm in 2 minds as to whether to add extra complexity to deal with this: should we automatically populate missing fields with zeros, or error out on the basis that these files do not comply with the specification in the SOLPS manual for fort.44?

I don't think that processing the case when the number of molecules `nm` is 0 adds any complexity, you can just add `if` statement each time you assign the molecular data:
```python
if eirene.nm:
    eirene.dm = raw_data['dmb2'].reshape(eirene.nm, eirene.ny, eirene.nx)
```
Actually, the same should be done for the parser of unlabelled files. I forgot that there are cases without molecules.

cherab/solps/eirene/parser/fort44_post_2016.py
vsnever: I asked Andrey Pshenov about this. "Total H-alpha emissivity" means the sum of ionization and recombination channels, so `emiss` is an atomic H-alpha emissivity, and this code is correct.

cherab/solps/eirene/parser/fort44_post_2016.py
jacklovell: OK, I'll modify it to deal with possibly being resolved over atoms too. I think it would be better to present a consistent interface in the `Eirene` object though: the shape of `Eirene.eradt` should not depend on whether the raw data is resolved or not. I propose we always provide `eradt` summed over all atoms, with shape `(1, ny, nx)`: this matches the existing behaviour with the original 2013 parser and is also consistent with the `emist` and `emism` data.

jacklovell:
> > I also tested it on a WEST simulation which only included He atoms, and it failed loudly because of missing dmb2 etc. entries: I think this is a good thing compared with silently reading garbage. I'm in 2 minds as to whether to add extra complexity to deal with this: should we automatically populate missing fields with zeros, or error out on the basis that these files do not comply with the specification in the SOLPS manual for fort.44?
> 
> I don't think that processing the case when the number of molecules `nm` is 0 adds any complexity, you can just add `if` statement each time you assign the molecular data:
> 
> ```python
> if eirene.nm:
>     eirene.dm = raw_data['dmb2'].reshape(eirene.nm, eirene.ny, eirene.nx)
> ```
> 
> Actually, the same should be done for the parser of unlabelled files. I forgot that there are cases without molecules.

Is it also possible to have cases without atoms or ions too? Perhaps all assignments should be checked with `if eirene.na`, `if eirene.nm` or `if eirene.ni` as appropriate.

vsnever:
> Is it also possible to have cases without atoms or ions too? Perhaps all assignments should be checked with `if eirene.na`, `if eirene.nm` or `if eirene.ni` as appropriate.

I thought about this too. Even if a simulation without atoms or ions is technically possible, it will not be a plasma simulation. So I do not think we should care about these cases because they are too exotic (if even exist).

jacklovell:
> > Is it also possible to have cases without atoms or ions too? Perhaps all assignments should be checked with `if eirene.na`, `if eirene.nm` or `if eirene.ni` as appropriate.
> 
> I thought about this too. Even if a simulation without atoms or ions is technically possible, it will not be a plasma simulation. So I do not think we should care about these cases because they are too exotic (if even exist).

I've found a run which just uses He: the fort.44 file has na=1, nm=0, ni=0. There will still be He-II and He-III in the B2 plasma though, just like in the generomak example the plasma has all 7 carbon ionisation stages even though the fort.44 file does not list C among the ions (only H2+ is included in fort.44). 

So it's perhaps not unreasonable to have these fort.44s without any molecules or ions: molecules may simply not be present in the SOLPS simulation at all, while the only ionised species will be in the B2 data. And B2 treats ions self-consistently, whereas its fluid-neutrals model is not self consistent which is why we need the Eirene data in the first place.

vsnever:
> I've found a run which just uses He: the fort.44 file has na=1, nm=0, ni=0. There will still be He-II and He-III in the B2 plasma though, just like in the generomak example the plasma has all 7 carbon ionisation stages even though the fort.44 file does not list C among the ions (only H2+ is included in fort.44).
> 
> So it's perhaps not unreasonable to have these fort.44s without any molecules or ions: molecules may simply not be present in the SOLPS simulation at all, while the only ionised species will be in the B2 data. And B2 treats ions self-consistently, whereas its fluid-neutrals model is not self consistent which is why we need the Eirene data in the first place.

I see. In terms of Eirene, the ions are the molecular ions. I forgot about that, sorry. So, yes, checking for `ni == 0` is necessary, it's not an exotic case at all.

jacklovell:
> > Is it also possible to have cases without atoms or ions too? Perhaps all assignments should be checked with `if eirene.na`, `if eirene.nm` or `if eirene.ni` as appropriate.
> 
> I thought about this too. Even if a simulation without atoms or ions is technically possible, it will not be a plasma simulation. So I do not think we should care about these cases because they are too exotic (if even exist).

I've found a run which just uses He: the fort.44 file has na=1, nm=0, ni=0. There will still be He-II and He-III in the B2 plasma though, just like in the generomak example the plasma has all 7 carbon ionisation stages even though the fort.44 file does not list C among the ions (only H2+ is included in fort.44). 

So it's perhaps not unreasonable to have these fort.44s without any molecules or ions: molecules may simply not be present in the SOLPS simulation at all, while the only ionised species will be in the B2 data. And B2 treats ions self-consistently, whereas its fluid-neutrals model is not self consistent which is why we need the Eirene data in the first place.

jacklovell:
I've pushed a couple more changes, and now the code can deal with simulations with no molecules or ions. I've tested it on a He-only WEST simulation and it runs without error and produces reasonable-looking plots.

cherab/solps/eirene/parser/fort44_post_2016.py
jacklovell: @vsnever do you know which SOLPS/Eirene combinations result in eneutrad being summed over the atoms? If it's only going to be the case for pre-2016 (i.e. unlabelled) fort.44s then I think we can simplify this parser and move the check into the other parsers. Reducing the number of branches in each parser is a major advantage of using different parsers for different fort.44 versions.

vsnever: I'm afraid, this is more complicated than just a version of the fort.44:
```fortran
        if (jvft44 .ge. 20071209) then

          if (streql(solps_version,'4.3').or.streql(solps_version,'5.1').or.streql(eirene_format,'facelift')) then
C           eneutrad data is summed over species!
            call gfsub2(44,ndx,ndy,nnx,nny,eneutrad(:,:,1,0))
          else
C           eneutrad data is resolved over species!
            do is = 1, natmi
              call gfsub2(44,ndx,ndy,nnx,nny,eneutrad(:,:,is,0))
            end do
          end if

        end if
```
This is from b2yt_ngread.F. I don't know if a combination of post-2016 fort.44 and `solps_version` < 5.2 is possible or not, but I saw references to the latest (2019) fort.44 version even in the source code of SOLPS 4.3.3.

vsnever:
Just tested this parser against the entire set of SOLPS-ITER example cases, and it works fine with all of them. However, in all these cases, the fort.44 files have the same version, 20170328.

jacklovell:
OK, great. Once @Mateasek has confirmed he's happy with it I'll merge.

Mateasek:
Hi @jacklovell and @vsnever , sorry for my late reply. I forgot about this PR. I looked into it and I have no suggestions, I think your knowledge about the sopls fomats is deeper than mine anyway. I tested loading of 20180323 and 20170328 fort.44 files and it seems it loaded both correctly, however it is difficult for me to check thoroughly. If there are no other matters you would like me to check with these two versions I think this PR should be accepted.

Mateasek (approved):
No requirements for changes from my side.

PR #42, opened 2020-12-24
Fix loading of total radiated power from raw files

### LINKED ISSUE #28: Inconsistent treatment of ENEUTRAD among input formats

jacklovell:
Load Eirene total radiated power in W/m3, where positive values
correspond to emission of radiation. Add line emission and
Bremsstrahlung from the plasma solution using the b2fplasmf file:
without this a significant fraction of the radiation is omitted. This
does mean a b2fplasmf file is required: if this is not available then
no value is assigned to SOLPSSimulation.total_radiation, as using the
Eirene neutral data alone would be misleading.

With this change, the SOLPSSimulation.total_radiation quantity is
comparable for simulations loaded from raw files, MDSplus or a
balance.nc file.

Fixes #28

### DIFF
--- a/cherab/solps/b2/parse_b2_block_file.py
+++ b/cherab/solps/b2/parse_b2_block_file.py
@@ -28,14 +28,22 @@ MESH_DATA = 1
 
 # Code based on a script by Felix Reimold (2016)
 
-def load_b2f_file(filepath, debug=False):
+def load_b2f_file(filepath, debug=False, header_dict=None):
     """
-    File for parsing the 'b2fstate' B2 Eirene output file.
+    File for parsing the 'b2fstate and b2fplasmf' B2 Eirene output files.
 
     :param str filepath: full path for file to load and parse
     :param bool debug: flag for displaying textual debugging information.
-    :return: tuple of dictionaries. First is the header information such as the version, label, grid size, etc. Second
-    dictionary has a SOLPSData object for each piece of data found in the file.
+    :param dict header_dict: header information
+    :return: tuple of dictionaries. First is the header information such
+      as the version, label, grid size, etc. Second contains arrays of
+      simulation outputs. Third contains information about the simulation
+      itself.
+
+    If header_dict is provided, it should contain the dimension info 'nx',
+    'ny', 'nxg', 'nyg', 'nxy', 'nxyg', as obtained by reading a previous
+    b2f* file. If header_dict is None, these values will be read from the
+    top of the file.
     """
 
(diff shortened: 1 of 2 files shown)


vsnever:
Great that you added the parsing of b2fplasmf file. I tried to convert the binary b2plasma files from SOLPS-ITER SDK examples to the formatted b2plasmf files using the b2uf utility, with no success. Did you manage to get b2uf working, or do you have the b2fplasmf files from someone else's SOLPS runs?

jacklovell:
I did manage to get b2uf working, but on some other users' runs rather than the examples. These runs had the binary b2plasma file but not b2plasmf. I had to use tcsh (couldn't get ksh to work) as follows:
```tcsh
cd <path to SOLPS code repository>
source setup.csh
cd <path to run>
b2run b2uf
```
I did find that in some cases it would run the entire simulation again, but they didn't take too long.

jacklovell:
Since there have been no objections, I'll merge this.

PR #43, opened 2021-04-01
Use generomak first wall for imaging plasma demo script

jacklovell:
Now that the mesh for the first wall is in core/development, use it to make the
demo where the SOLPS simulation is imaged with a camera into a more
compelling demonstration.

### DIFF
--- a/demos/generomak_raw_files_image_plasma.py
+++ b/demos/generomak_raw_files_image_plasma.py
@@ -18,7 +18,7 @@
 import os
 import matplotlib.pyplot as plt
 
-from raysect.core.math import Vector3D, translate, rotate_basis
+from raysect.core.math import Vector3D, translate, rotate_basis, rotate_z
 from raysect.optical import World
 from raysect.optical.observer import PinholeCamera, PowerPipeline2D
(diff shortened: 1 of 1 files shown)


vsnever:
I rendered this at higher resolution and with better statistics. Looks nice!
generomak_raw_files_image_plasma

jacklovell:
Yes, it really does produce a nice image with higher resolution and more rays. Unfortunately I think the run time to produce an image of this quality wouldn't be suitable for a demo, but perhaps we can use this illustration in the HTML documentation once that starts to include generomak components.

vsnever:
To render this image, I pre-calculated the total radiation on the SOLPS grid and then used the `RadiationFunction` emitter. The rendering took ~20 min, but we should leave this for a separate demo.

Mateasek (approved):
Fine by me. This will provide awesome pictures

Mateasek:
Just to the high-resolution case, users can simply increase the number of pixels and use the same demo to get a better resolution image. I don't think there is a need for 2 separate demos. The image @vsnever generated can be used for illustrations as @jacklovell said.

PR #44, opened 2021-04-01
Fix deprecation and syntax warnings in recent Python and Numpy

jacklovell:
Numpy 1.20 deprecated np.int, np.bool etc.
Python 3.8 now raises a SyntaxWarning when using is with literals.

### DIFF
--- a/cherab/solps/formats/balance.py
+++ b/cherab/solps/formats/balance.py
@@ -45,8 +45,8 @@ def load_solps_from_balance(balance_filename):
 
         species_list = []
         neutral_indx = []
-        am = np.round(fhandle.variables['am'].data).astype(np.int)  # Atomic mass number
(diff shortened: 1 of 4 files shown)


cherab/solps/formats/balance.py
```diff
@@ -94,7 +94,7 @@ def load_solps_from_balance(balance_filename):
             for key in fhandle.variables.keys():
                 if 'fna_' in key:
                     fna += fhandle.variables[key].data.copy()
-            if fna is not 0:
+            if fna != 0:
```
vsnever: Please, replace with:
```
if np.any(fna != 0):
```

jacklovell: Yep, just saw this when I tried to run these examples. Thanks.

vsnever:
I ran the tests. Everything works fine, so I think it's safe to merge.

PR #49, opened 2021-05-22
Add H-alpha emissivity calculated with EIRENE to SOLPSSimulation.

### LINKED ISSUE #48: Add molecular and total H-alpha emission to SOLPSSimulation.
EIRENE can calculate H-alpha emission due to atomic and molecular processes. Here, molecular processes are the dissociation of hydrogen molecules and molecular ions with the production of atoms in excited states. Such atoms are not taken into account in the ADAS collisional radiative model. Moreover, the atomic-only component of H-alpha emission calculated with EIRENE is usually 1.5 - 2 times higher (at least in the EIRENE versions for SOLPS) than the excitation + recombination emission calculated with ADAS. This difference is likely due to the difference in hydrogen data in EIRENE and ADAS.

I would like to have the molecular-only and total (atomic + molecular) H-alpha emissivities as `SOLPSSimulation` attributes, so that  they can be used to calculate H-alpha intensities bypassing the standard Cherab methods, which are based on the ADAS data.

Since I needed this feature for my work, I already implemented it in a dedicated branch. I'll make a PR, so you can view the changes.

vsnever:
This PR adds new attributes to `SOLPSSimulation`, which are used to store the molecular and total (atomic + molecular) H-alpha emissivities calculated with EIRENE. Like all other simulation data, the emissivities are also available as `Function2D` and `Function3D` objects. This PR fixes #48.

@Mateasek, @jacklovell, what do you think on this?


cherab/solps/formats/raw_simulation_files.py
```diff
@@ -168,6 +170,19 @@ def load_solps_from_raw_output(simulation_path, debug=False):
             total_radiation[1:-1, 1:-1] -= eradt_raw_data
             sim.total_radiation = total_radiation / mesh.vol
 
+        # Obtaining molecular and total H-alpha radiation
+        halpha_wavelength = OpenADAS().wavelength(hydrogen, 0, (3, 2))
```
Mateasek: I think it would be better (correct and clearer) if this respected isotopes. I know that it will not change the radiated power much and is negligible, but since the issue with the isotope wavelength is being fixed ;-)

Mateasek (changes requested):

cherab/solps/formats/raw_simulation_files.py
vsnever: I agree, however this might be tricky in the case of multi-isotope simulation. EIRENE sums H-alpha emission over all isotopes, therefore the effective Balmer-alpha photon energy must be calculated with respect to local isotope concentrations. I'll try to do this photon to energy conversion more accurately.

vsnever: I rewrote the photon-to-energy conversion taking into account the isotopic composition. The new code will give an accurate result in the case of single isotope simulations (probably over 99% of all SOLPS simulations), as well as for the impact excitation part of the Balmer-alpha emissivity in the case of multi-isotope simulations. I think that a more accurate conversion is impractical, since it will require atomic and molecular data used in SOLPS for this particular simulation.

jacklovell:
I'm happy with this PR. @Mateasek are you happy for it to be merged too?

Mateasek:
Sorry for the delay, no suggestions from my side. Thanks for the work @vsnever

PR #51, opened 2021-05-27
fix: use plot_triangle_mesh instead of plot_mesh

Mateasek:


Mateasek:
@vsnever we will also need to update the docs: https://github.com/cherab/core/blob/master/docs/source/demonstrations/solps/solps_plasma.rst

vsnever:
Thanks! Now everything is fine.

Yes, these docs should be updated as well.

jacklovell:
That demo in the Core docs is out of date anyway: it's part of the Code Examples Gallery which needs to be cleaned up and moved. It would be better to move it over to cherab-solps and start a documentation set within this repository for SOLPS-specific demos.

PR #52, opened 2021-07-03
Replace solps_total_radiated_power() with make_solps_emitter() and remove SOLPSTotalRadiatedPower

### LINKED ISSUE #50: Fix SOLPSTotalRadiatedPower and make solps_total_radiated_power() more universal
When calculating the spectral emissivity in `SOLPSTotalRadiatedPower.emission_function()`, the total emissivity is divided by 4π, wavelength range and the number of spectral bins. It seems to me that dividing by the number of spectral bins is unnecessary here and breaks the normalisation.

I propose a change from:
```python
spectrum.samples_mv[:] = self.total_rad.evaluate(point.x, point.y, offset_z) / (4 * pi * wvl_range * spectrum.bins)
```
to:
```python
spectrum.samples_mv[:] = self.total_rad.evaluate(point.x, point.y, offset_z) / (4 * pi * wvl_range)
```
Note, however, that even after this correction, this emitter will not work properly if dispersive rendering is enabled.

I also propose to rename `SOLPSTotalRadiatedPower` to `SOLPSRadiatedPower`, `solps_total_radiated_power()` to `solps_radiated_power()` and convert the latter to a more universal function. The new function will look like this:
```python
def solps_radiated_power(world, solps_mesh, solps_radiated_power, step=0.01):

    if not isinstance(solps_mesh, SOLPSMesh):
        raise TypeError('Argument solps_mesh must be a SOLPSMesh instance.')
    if not isinstance(solps_radiated_power, SOLPSFunction2D):
        raise TypeError('Argument solps_radiated_power must be a SOLPSFunction2D instance.')

    radiated_power_f3d = AxisymmetricMapper(solps_radiated_power)
    outer_radius = solps_mesh.mesh_extent['maxr']
    inner_radius = solps_mesh.mesh_extent['minr'] - 0.001
    plasma_height = solps_mesh.mesh_extent['maxz'] - solps_mesh.mesh_extent['minz']
    lower_z = solps_mesh.mesh_extent['minz']

    main_plasma_cylinder = Cylinder(outer_radius, plasma_height, parent=world,
                                    material=SOLPSRadiatedPower(radiated_power_f3d, vertical_offset=lower_z, step=step),
                                    transform=translate(0, 0, lower_z))

    return main_plasma_cylinder
```

After this change, the `solps_radiated_power()` can be used for example to calculate the wavelength-integrated H-alpha signals (issue #48 and PR #49).

@Mateasek, @jacklovell, what do you think?

vsnever:
This PR replaces `solps_total_radiated_power()` with a more general `make_solps_emitter()` and removes redundant `SOLPSTotalRadiatedPower`, because in `make_solps_emitter()` a combination of `RadiationFunction` and `VolumeTransform` is used instead. A script that demonstrates the usage of `make_solps_emitter()` on the example of total radiated power is added.

This fixes #50.


demos/generomak_raw_files_solps_emitter.py
```diff
@@ -0,0 +1,103 @@
...
+    # Calculates total radiation power density on SOLPS mesh
+    direction = Vector3D(0, 0, 1)
+    total_radiation = np.zeros((mesh.ny, mesh.nx))
+    for model in plasma.models:
+        for ix in range(total_radiation.shape[1]):
+            for iy in range(total_radiation.shape[0]):
+                point = Point3D(mesh.cr[iy, ix], 0, mesh.cz[iy, ix])
+                total_radiation[iy, ix] += 4. * np.pi * model.emission(point, direction, spectrum.new_spectrum()).total()
+
+    return total_radiation
```
jacklovell: Since this function is only every called once in this demo, I think it makes more sense just to inline it. The narrative in the demo will flow better that way, otherwise users have to skip over this function when they first start reading the demo and then jump back up to it later on.

demos/generomak_raw_files_solps_emitter.py
```diff
@@ -0,0 +1,103 @@
...
+plasma.atomic_data = OpenADAS(permit_extrapolation=True)
+for species in plasma.composition:
+    if species.charge < species.element.atomic_number:
+        plasma.models.add(TotalRadiatedPower(species.element, species.charge))
+
+# Caching the total radiation power density on the SOLPS mesh.
+print('Caching total radiation power density on SOLPS mesh...')
+spectrum = Spectrum(camera.min_wavelength, camera.max_wavelength, camera.spectral_bins)
+total_radiation = total_radiation_on_mesh(sim.mesh, plasma, spectrum)
+sim.total_radiation = total_radiation  # this creates sim.total_radiation_f2d and sim.total_radiation_f3d
```
jacklovell: I would move this above the creation of the camera, as it's more closely related to building the SOLPSSimullation / plasma object than it is actually doing the imaging of said object.

jacklovell (changes requested):
This looks good to me @vsnever. I only have a couple of very minor changes to request in the demo script to make it a bit easier to follow. The way you've implemented things I think is good.

It's interesting that there's such a large performance improvement doing it this way: I get a factor ~40 speedup compared with the simpler "image plasma" demo. The same models are being evaluated in both cases, but I expect this demo uses more cached evaluations of the model because the AxisymmetricMapper doesn't cache values (x, y, z) which map to the same (r, z).

vsnever:
Hi @jacklovell, thanks for the review. I improved the demo as you suggested. Also, I set the `display_unsaturated_fraction` parameter of `PowerPipeline2D` to 0.9999 to prevent underexposure that can occur in case of poor ray statistics.

demos/generomak_raw_files_solps_emitter.py
vsnever: Done.

demos/generomak_raw_files_solps_emitter.py
vsnever: Done.

jacklovell (approved):

PR #56, opened 2021-07-05
Make SOLPS line emitter a PlasmaModel with user-provided emissivity

### LINKED ISSUE #55: Fix BasicSolpsLineEmitter and make it a PlasmaModel subclass
`BasicSolpsLineEmitter` is also affected by the issue cherab/core#235. Namely, electron temperature is used for thermal broadening of the line instead of species temperature. I suggest making SOLPS line emitter a subclass of `PlasmaModel` with user-provided emissivity defined as a `SOLPSFunction2D`. When converted to `PlasmaModel`, it will support any line shape model available in Cherab, not just `GaussianLine`.

This emitter might be useful when #49 is merged.

I'll make a draft PR to show the changes I propose.

vsnever:
This fixes #55 by converting SOLPS line emitter to a `PlasmaModel`. With this change, it supports any line shape model available in Cherab. 
Also, `BasicSolpsLineEmitter` is renamed to `SOLPSLineEmitter`.


vsnever:
I've added a demo, so this PR is now ready for review. The patch cherab/core#323 must be applied first otherwise the demo may not work correctly.

cherab/solps/models/line_emitter.pyx
```diff
@@ -0,0 +1,111 @@
...
+
+        # add emission line to spectrum
+        radiance = RECIP_4_PI * self._power_density(point.x, point.y, point.z)
+        return self._lineshape.add_line(radiance, point, direction, spectrum)
+
+    cdef int _populate_cache(self) except -1:
+
+        # sanity checks
+        if self._plasma is None or self._atomic_data is None:
+            raise RuntimeError("The emission model is not connected to a plasma object.")
```
jacklovell: Separate this into two checks. Otherwise if a user manages to supply a plasma but no atomic data they will get a misleading error message.

vsnever: I fixed this and also checked other plasma models from Cherab/core. The `ExcitationLine`, `RecombinationLine`, `TotalRadiatedPower` and `BeamCXLine` models are affected by the same issue. I'll rise the issue and prepare a fix in Cherab/core.

jacklovell (approved):
This looks good to me. Thanks for providing the demo too.

jacklovell:
Happy to merge after @Mateasek reviews.

cherab/solps/models/line_emitter.pyx
```diff
@@ -0,0 +1,114 @@
...
+
+        # locate target species
+        try:
+            self._target_species = self._plasma.composition.get(self._line.element, self._line.charge)
+        except ValueError:
+            raise RuntimeError("The plasma object does not contain the ion species for the specified line "
+                               "(element={}, ionisation={}).".format(self._line.element.symbol, self._line.charge))
+
+        # identify wavelength
+        self._wavelength = self._plasma.atomic_data.wavelength(self._line.element, self._line.charge, self._line.transition)
```
Mateasek: Hi @vsnever , thanks for this PR. I have only one question. Shouldn't the wavelength be obtained from self._atomic_data instead? If the user specifies the atomic_data source, it should take the values from there in my opinion, instead from the plasma. Or you could make it optional and specify in the docstring, that if there is no atomic_data source specified, it will take the one from plasma.

cherab/solps/models/line_emitter.pyx
```diff
@@ -0,0 +1,114 @@
...
+
+
+cdef class SOLPSLineEmitter(PlasmaModel):
+    """
+    Line emitter that uses a custom `SOLPSFunction2D` function for radiation power density.
+
+    :param SOLPSFunction2D radiation_function: The power density function defined on the SOLPS mesh.
+    :param Line line: Spectral line.
+    :param Plasma plasma: Plasma object. Default is None.
+    :param AtomicData atomic_data: Atomic data provider. Will use `OpenADAS` if None (default).
```
Mateasek, line 37: I can't find where you specify the OpenADAS() when None is passed, is this accurate?

vsnever: This is a good question. When the model is added to the `plasma.models`, the `atomic_data` of that plasma object will be used. This is invoked in `Plasma._configure_geometry()` and is done in `PlasmaMaterial` initialiser:
https://github.com/cherab/core/blob/0eb7454137b17bfbeca6d5e3eae4281f9cd4c157/cherab/core/plasma/material.pyx#L25-L46. 
So, it does not matter what values for `plasma` and `atomic_data` are provided at initialisation, they will be overwritten, as soon as the model is added to `Plasma.models`. The only way to use the `atomic_data` different from that of the plasma object, is by setting `atomic_data` once the model is already added to the `plasma.models` list. However, one can use the `PlasmaModel` without adding it to the `models` list for example to sample the emission spectrum, and this can be done with the `atomic_data` provided at initialisation.

The `atomic_data` is not set to `OpenADAS()` by default in either `Plasma` or any implemented `PlasmaModel`. I am just following this pattern here. I suspect this goes back to the days when `cherab.openadas` was distributed separately from Cherab, and now this can be fixed. But, if we are going to fix this, I think it should be done in the parent class.

cherab/solps/models/line_emitter.pyx
vsnever: Thanks, @Mateasek, you are right. The wavelength should be always obtained from `self._atomic_data` just like in other plasma models. I'll fix this.

vsnever: Done.

cherab/solps/models/line_emitter.pyx
Mateasek: Right, sorry I missed this. Thanks for pointing it out @vsnever

cherab/solps/models/line_emitter.pyx
Mateasek: thanks

Mateasek (approved):

Mateasek:
I think this can be accepted @jacklovell

PR #61, opened 2021-07-29
Fix SOLPSMesh.vol and some other docstrings in SOLPSMesh.

### LINKED ISSUE #60: Improve docstring of sim.mesh.vol
When loading a simulation using `sim = load_solps_from_raw_output(path)`, the docstring of `sim.mesh.vol` reads "Volume/area of each grid cell". It would be more instructive to make it "Volume of each grid cell", with the unit specified as well.

vsnever:
This fixes #60 and some other docstrings in `SOLPSMesh`.

### DIFF
--- a/cherab/solps/mesh_geometry.py
+++ b/cherab/solps/mesh_geometry.py
@@ -27,16 +27,16 @@ class SOLPSMesh:
     """
     SOLPSMesh geometry object.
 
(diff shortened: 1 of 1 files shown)


Mateasek:
I think this is ready to be merged, thanks for the quick reaction @vsnever

jacklovell:
Fine with me. Thanks @vsnever.

PR #62, opened 2021-07-29
Allow the user to specify individual paths to SOLPS raw ouput files in load_solps_from_raw_output()

### LINKED ISSUE #59: Loading SOLPS-ITER simulation: refer to baserun for b2fgmtry
When using `load_solps_from_raw_output`, the `b2fgmtry` file is expected to be in the run directory (which is passed as argument to `load_solps_from_raw_output`). However, usually one sets up the geometry using `b2run b2ag` in the baserun directory  (usually `../baserun` relative to the run directory), so `b2fgmtry` is stored there and SOLPS refers to it from every run directory. It would, therefore, be helpful if Cherab searched for `b2fgmtry` in baserun if the file isn't found in the run directory.

vsnever:
~~This fixes #59 by adding `baserun_path` argument to `load_solps_from_raw_output()`. Now if the SOLPS file is not found in `simulation_path`, the version from `baserun_path` will be used.~~

This fixes #59 by adding `mesh_file_path`, `b2_state_file_path`, `b2_plasma_file_path` and `eirene_fort44_file_path` arguments to `load_solps_from_raw_output()`. Now the user can specify the path to any of the SOLPS output file, if this file is not in the `simulation_path` directory.

### DIFF
--- a/cherab/solps/formats/raw_simulation_files.py
+++ b/cherab/solps/formats/raw_simulation_files.py
@@ -30,7 +30,7 @@ from cherab.solps.solps_plasma import SOLPSSimulation, prefer_element, eirene_fl
 
 
 # Code based on script by Felix Reimold (2016)
-def load_solps_from_raw_output(simulation_path, baserun_path=None, debug=False):
+def load_solps_from_raw_output(simulation_path, debug=False, baserun_path=None):
     """
     Load a SOLPS simulation from raw SOLPS output files.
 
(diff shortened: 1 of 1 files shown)


cherab/solps/formats/raw_simulation_files.py
```diff
@@ -30,7 +30,7 @@
 
 
 # Code based on script by Felix Reimold (2016)
-def load_solps_from_raw_output(simulation_path, debug=False):
+def load_solps_from_raw_output(simulation_path, baserun_path=None, debug=False):
```
jacklovell: This will break existing code which passes `debug` as a positional argument. Please put `baserun_path` as  the third argument instead.

vsnever: Done.

Mateasek:
Thanks for the PR @vsnever

jacklovell:
Looks good to me. Thanks @vsnever

jacklovell (approved):

PR #64, opened 2021-08-09
Fix an error in SOLPSSimulation.electron_velocities_cylindrical setter

### LINKED ISSUE #63: Fix SOLPSSimulation.electron_velocities_cylindrical setter
There is a copy paste error in `SOLPSSimulation.electron_velocities_cylindrical` setter:
https://github.com/cherab/solps/blob/a4e6b63b45654d7bf438a27eb47b92f544474b67/cherab/solps/solps_plasma.py#L310-L318
Instead of this:
```python
velocities[:, 2] = -value[:, 1]
```
should be this:
```python
velocities[2] = -value[1]
```
I think we need unit tests to track bugs like this. I'll start adding them.

vsnever:
This fixes #63 and adds a unit test for vector transforms from poloidal to cylindrical coordinates and vice versa performed in the `SOLPSSimulation` setters for the vector values.

Other unit tests can be added to `TestSOLPSSimulation` in the future.

### DIFF
--- a/cherab/solps/tests/test_simulation.py
+++ b/cherab/solps/tests/test_simulation.py
@@ -13,54 +13,65 @@ class TestSOLPSSimulation(unittest.TestCase):
 
     def setUp(self):
         # Creating a test SOLPSMesh
-        rv = np.linspace(1., 2., 11)
-        zv = np.linspace(-1., 1., 21)
-        r2d, z2d = np.meshgrid(rv, zv, indexing='ij')
-        vol = (np.pi * (rv[1:] + rv[:-1]) * np.diff(rv))[:, None] * np.diff(zv)[None, :]
-        ny = rv.size - 1
-        nx = zv.size - 1
+        # closed magnetic surfaces only with circular cross section
+        r0 = 2.  # major radius
+        beta = np.linspace(0, 2 * np.pi, 19)  # poloidal angle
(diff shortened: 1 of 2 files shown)


cherab/solps/tests/test_simulation.py
```diff
@@ -0,0 +1,99 @@
...
+        neighbiy[0, 1:, :] = np.arange(ny - 1, dtype=int)[:, None]  # left
+        neighbiy[1, :, :] = np.arange(ny, dtype=int)[:, None]  # bottom
+        neighbiy[2, :-1, :] = np.arange(1, ny, dtype=int)[:, None]  # right
+        neighbiy[3, :, :] = np.arange(ny, dtype=int)[:, None]  # top
+        mesh = SOLPSMesh(r, z, vol, neighbix, neighbiy)
+
+        # Creating a test SOLPSSimulation
+        species_list = [('hydrogen', 0), ('hydrogen', 1)]
+        self.sim = SOLPSSimulation(mesh, species_list)
+        rng = default_rng()
```
jacklovell: I'd suggest using a fixed seed for the random number generator. That way the tests will be repeatable, and allow us to check against known-good values.

vsnever: I agree, now the seed is fixed.

PR #66, opened 2021-11-25
Allow user-provided AtomicData in load_solps_from_mdsplus() and load_solps_from_raw_output().

### LINKED ISSUE #65: Loading simulation from MDSPlus or raw files will fail at obtaining Balmer-alpha wavelengths if OpenADAS repository is not installed in the default location
When converting Balmer-alpha radiation density from photons/s/m3 to W/m3, the OpenADAS repository is used to obtain the wavelengths of Balmer-alpha spectral lines: https://github.com/cherab/solps/blob/31ec6b5db25be9f31c9e7a992845ccbbf58e6919/cherab/solps/formats/raw_simulation_files.py#L191-L197

https://github.com/cherab/solps/blob/31ec6b5db25be9f31c9e7a992845ccbbf58e6919/cherab/solps/formats/mdsplus.py#L182-L188

If the OpenADAS data path is not the default one, the code will fail to locate the wavelength files. 

Instead of adding optional `atomic_data` argument to `load_solps_from_raw_output()` and `load_mesh_from_mdsplus()`, I propose to define a constant:
```python
BALMER_ALPHA_WAVELENGTH = {hydrogen: 656.279, deuterium: 656.101, tritium: 656.041}
```
The value for hydrogen is hardcoded in `populate()` anyway and the other two wavelengths are calculated using the hydrogen one. 

Does anyone have any objection to hardcoding these values here?

vsnever:
This fixes #65 by allowing the user to provide a custom `AtomicData` instant as a parameter to `load_solps_from_mdsplus()` and `load_solps_from_raw_output()`. The `AtomicData` is used in these functions to convert the radiation density from photons/s/m<sup>3</sup> to W/m<sup>3</sup>.

### DIFF
--- a/cherab/solps/formats/mdsplus.py
+++ b/cherab/solps/formats/mdsplus.py
@@ -27,12 +27,15 @@ from cherab.solps.mesh_geometry import SOLPSMesh
 from cherab.solps.solps_plasma import SOLPSSimulation, prefer_element, eirene_flux_to_velocity, b2_flux_to_velocity
 
 
-def load_solps_from_mdsplus(mds_server, ref_number):
+def load_solps_from_mdsplus(mds_server, ref_number, atomic_data=None):
     """
     Load a SOLPS simulation from a MDSplus server.
 
     :param str mds_server: Server address.
     :param int ref_number: Simulation reference number.
+    :param AtomicData atomic_data: The atomic data provider. Used here to convert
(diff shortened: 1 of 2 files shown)


Mateasek (approved):
Thanks for adding the possibility to pass custom atomic_data repository. The PR seems good to me.

PR #68, opened 2021-12-21
Release 1.2.0

jacklovell:
Now that Cherab 1.3.0 has been released, we should release an updated cherab-solps module that is compatible with it.


jacklovell:
Release candidate packages are available on PyPI at https://pypi.org/project/cherab-solps/. Manylinux wheels are provided for Python 3.7, 3.8 and 3.9, and there's an sdist for other Python versions. Please test these and report any problems in this PR.

PR #74, opened 2024-08-28
Update for compatibility with Cherab core 1.5.

jacklovell:
I think this probably merits a minor version bump, despite there being no differences in the SOLPS-specific submodule, given the more significant changes in the core module.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ with open("README.md") as f:
 
 setup(
     name="cherab-solps",
-    version="1.2.1",
+    version="1.3.0.dev1",
     license="EUPL 1.1",
     namespace_packages=['cherab'],
     description="Cherab spectroscopy framework: SOLPS submodule",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -63,6 +63,6 @@
     long_description_content_type="text/markdown",
     packages=find_packages(),
     include_package_data=True,
-    install_requires=["raysect==0.7.1", "cherab==1.4"],
+    install_requires=["raysect==0.8.1.*", "cherab==1.5.*"],
```
vsnever, line 66: Thanks for this, @jacklovell. 
All is well, except do we actually need `raysect==0.8.1.*` in `install_requires` if we already have `cherab==1.5.*` which will install it anyway?

jacklovell: There are places in `cherab-solps` that directly use Raysect, e.g.: 
https://github.com/cherab/solps/blob/aecedf26bcbbf4121730692a10e488de5835bfa7/cherab/solps/solps_2d_functions.pyx#L24

I think it's good practice to explicitly state the dependency for this package rather than relying on another dependency package to satisfy it for us. Although in this case it's unlikely Cherab will drop Raysect (https://github.com/cherab/core/issues/445 notwithstanding) it's still a good habit to get in to.

vsnever: Thanks for the explanation, I agree.

vsnever (approved):

PR #77, opened 2024-08-28
Add demos installation to setup.py

### LINKED ISSUE #73: Package demo scripts
The core module now installs demo scripts in `share/cherab/demos/core`. I think, `cherab-solps` should do the same and install the demos in `share/cherab/demos/solps`.

vsnever:
This fixes #73 by adding a demo installation block to `setup.py`. The demos are installed to `share/cherab/demos/solps`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,11 @@
+from collections import defaultdict
 from setuptools import setup, find_packages, Extension
 from Cython.Build import cythonize
 import sys
 import numpy
 import os
 import os.path as path
+from pathlib import Path
 
 force = False
 profile = False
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -65,4 +77,5 @@
     include_package_data=True,
```
jacklovell: I think it makes sense to unify the packaging approach we use in Cherab core. Remove MANIFEST.in and specify `package_data` in setup.py, like in https://github.com/cherab/core/blob/26595cb6101e14d6a3eeb1b7421024425114c7a1/setup.py#L130. For cherab-solps we only need the pyx and pxd files as there's no additional data in the module itself (only the demo data which is handled separately):

```
    packages=find_packages(),                                                                 
    package_data={"": [                                                                       
        "**/*.pyx", "**/*.pxd",  # Needed to build Cython extensions.                         
        ],                                                                                    
    },                                                                                        
    data_files=data_files,
```

vsnever: Yes, I agree. It's done.

jacklovell (approved):

PR #79, opened 2024-10-28
Build: Switch from setuptools to scikit-build-core

jacklovell:
Move away from the setuptools-based build procedure, and use a modern framework, scikit-build-core, which is PEP518-compliant and uses CMake and Ninja for building the cython extensions.

Notable improvements:

* Automatically rebuilds extensions when Cython sources change, when installed in editable mode.
* Parallel by default, doesn't require hacking setuptools to do so.
* Out of tree builds, keeps the source tree clean.
* Profiling (and line profiling) can be enabled for automatic rebuilds without having to specify `--profile` or `--line-profile` on the command line every time.
* No more deprecated calls to `python setup.py <command>`, e.g. in `./dev/build.sh`.

Other miscellanous changes:
* Use Ubuntu 22.04 in CI, since 24.04 no longer includes Python 3.7.
* CI on Python 3.7 through 3.12, which are supported by upstream Cherab and Raysect.

Couple of points of discussion before merging:
* The `--profile` and `--line-profile` replacements are cached, so will need to be unset before building release wheels. I've updated the dev/build_wheels script to handle this for Python 3.7-3.9, and it can be easily adapted for manylinux2014 for 3.10-3.12. So there will hopefully not be a risk of accidentally releasing wheels with profiling still enabled. But the other option is to use environment variables which can be set every time (see e.g. the `--annotate` option in the top level CMakeLists.txt). Is this worth considering instead (or as well, for certain options)?
* Replacing pkg-resouces style namespaces from setuptools with PEP420-style namespaces may have some implications for compatibility with other Cherab submodules. It looks like this isn't a problem with standard `pip install` since that uses PEP420 anyway, and I haven't noticed any problems with installing it in develop mode yet but haven't tested it exhaustively.


Mateasek (approved):
Hi @jacklovell ,

thank you very much for the contribution. I tested it since I run python 3.10 where the latest version of the package doesn't install. This branch installation went without any problems. Since what you've done here is way out of my scope of knowledge, I have no idea what else to check.

cherab/solps/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+cmake_path(RELATIVE_PATH CMAKE_CURRENT_BINARY_DIR BASE_DIRECTORY ${CMAKE_BINARY_DIR} OUTPUT_VARIABLE libdir)
+
+cython_transpile(solps_2d_functions.pyx LANGUAGE C OUTPUT_VARIABLE solps_2d_functions_c)
+python_add_library(solps_2d_functions MODULE "${solps_2d_functions_c}" WITH_SOABI)
+target_link_libraries(solps_2d_functions PRIVATE Python::NumPy)
+
+install(TARGETS solps_2d_functions DESTINATION ${libdir})
```
vsnever, line 7: Here we have only a single file to compile. However, if we have more files to compile in the future, we have to add these rules for each of them. It might be better to make this file universal already now to make potential new additions easier. Since the compilation rules are more or less universal, you can make a list of filenames and go through them with `foreach` loop. Maybe, we can also use `GLOB` to automatically add all .pyx files in the subdirectory, but I'm not sure if that's better or not.

vsnever (approved):
Hi @jacklovell,

Thanks for this PR. I haven't worked with cmake before, so I had to dig into it to make a review. I haven't checked the scikit cmake files as I assume they are fine. The rest is fine too, and I only have one comment on the `CMakeLists.txt` files in the subdirectories. Regardless, I think the PR can be approved as is.

cherab/solps/CMakeLists.txt
jacklovell: Yes, this is a distinct departure from the programmatic way of doing builds with a `setup.py` file, in which you can write procedural code to find all applicable source files to be compiled. But having done some research online it appears this is now the recommended way to use build systems.

The related meson project has some discussion on this, and their position (which I'm coming round to) is that it is better to be explicit than implicit about what is compiled, where it's compiled and how it's compiled. They reason that a build system shouldn't be, or at least shouldn't be used as, Turing-complete because this tends in the long run to introduce excessive complexity and lead to long term problems at the expense of short term time saves.

I can see where they're coming from here: the previous `setup.py` required staring at for some time to work out exactly what it was doing and why. At least these CMakeLists.txt files are pretty easy to interpret, even for those with no CMake experience. There's some up-front effort to add a new file (which I've tried to address with instructions in the README), but Cython files are added less frequently than they're modified, so the net impact of requiring users to add the new file should be small. And it'll be very obvious if it hasn't been done, as the new module won't be importable at all.

jacklovell:
Thanks both. Do you have any thoughts on the first discussion bullet point above? Having had some time to reflect on this I think I'll remove the possibility of using environment variables to set build options, and keep only the CMake defines which will be cached across builds. I think the use of `--fresh` when building release wheels is sufficient to allay my original concerns about accidentally leaking profiling or debug build options into a release.

cherab/solps/CMakeLists.txt
vsnever: Thanks for the explanation. I understand this logic. I also explicitly set compilation rules for each file in my C++ projects.

vsnever:
> Thanks both. Do you have any thoughts on the first discussion bullet point above? Having had some time to reflect on this I think I'll remove the possibility of using environment variables to set build options, and keep only the CMake defines which will be cached across builds. I think the use of `--fresh` when building release wheels is sufficient to allay my original concerns about accidentally leaking profiling or debug build options into a release.

I agree. I don't really like specifying compiler directives via environment variables because these variables are used to configure the build or runtime environment. Compiler directives don't seem to be environment related to me.

Mateasek:
As I said, I have negligible experience with cmake or with compilation in general. Using environment variables can lead to some painful debugging when working across different environments and institutes with different IT approaches (personal experience). 

Do I understand it correctly that if we avoid environment variables, we don't prevent anything, we just require users to change the CMake files in order to achieve some change? If this is the case then what you suggest is in my opinion a better and cleaner approach:

-  Clear, defined and tracked state of the default compilation setup. 
- Easier for developers to maintain.
- Anyone can play around if they want to.

I hope I'm not writing a complete rubbish here. If so, please just ignore me.

vsnever:
> Do I understand it correctly that if we avoid environment variables, we don't prevent anything, we just require users to change the CMake files in order to achieve some change?

Changing the CMake files is not required in any case. Currently, to enable Cython function profiling, the user needs to run the install with `pip install --config-settings=cmake.define.profile=ON`. The problem is that CMake caches the definitions. And if the user re-runs the install in the same environment, just with `pip install`, profiling will still be enabled until the user explicitly runs `pip install --config-settings=cmake.define.profile=OFF`. For users unfamiliar with this, this behaviour may come as a surprise. @jacklovell suggested using environment variables instead of CMake definitions to control build options, since even though they behave the same as cached definitions (build option remains active until the value of the respective variable is explicitly changed), that behaviour is expected.

Mateasek:
Thanks for explanation @vsnever . I still think using environment variables is not a good option. Most of the users will not have to play with the cmake settings. Those who will are probably developing the code and such users can be expected to learn this slight overhead. For example, we can put a note somewhere in the documentation how to switch the flags and that they need to be switched back.

vsnever:
> Thanks for explanation @vsnever . I still think using environment variables is not a good option. Most of the users will not have to play with the cmake settings. Those who will are probably developing the code and such users can be expected to learn this slight overhead. For example, we can put a note somewhere in the documentation how to switch the flags and that they need to be switched back.

I think so too. I just tried to explain @jacklovell proposal as I understood it.

jacklovell:
OK, thanks for the feedback. I've removed support for environment variables. Also updated the Changelog, since I realised I forgot to do this before.

cherab/solps/CMakeLists.txt
eli-schwartz: In fact, even cmake warns against globbing in the documentation for that globbing support itself:
https://cmake.org/cmake/help/latest/command/file.html#glob



Note

> NOTE: We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The `CONFIGURE_DEPENDS` flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if `CONFIGURE_DEPENDS` works reliably, there is still a cost to perform the check on every rebuild.

Meson doesn't support globbing at all. It makes up for it in some other ways, for example, by natively supporting cython:

```meson
py = import('python').find_installation()
numpy_dep = dependency('numpy')

py.extension_module('line_emitter', 'line_emitter.pyx',
    dependencies: [numpy],
    install: true,
    subdir: 'cherab/solps/models',
)
```

